PR #18, opened 2026-05-15
Add mTLS support

### LINKED ISSUE #2: Add mTLS support
For supporting client certificate authentication.

HTTP client considerations:

* bundle downloads
* discovery (not supported)
* `http.send` built-in (not supported)

sspaink:
resolve: https://github.com/open-policy-agent/java-opa-sdk/issues/2

Adds mTLS support for service-to-server traffic (bundle downloads, decision-log uploads, status reports, discovery). Mirrors Go-OPA's `services.<name>.tls` and `services.<name>.credentials.client_tls` blocks so existing OPA configs port over.

Additional changes:
  - Bundle polling now uses jittered `[min, max]` delay (matching Go-OPA) instead of the prior `scheduleAtFixedRate(maxDelay)`. 
  - `BundlePlugin.validate` now flags references to unknown services up front instead of only logging on first download.
  
    ## Config additions

  ```yaml
  services:
    acmecorp:
      url: https://policy.example.com
      tls:
        ca_cert: /etc/ssl/corp-ca.pem
        system_ca_required: true        # also trust JVM defaults
      credentials:
        client_tls:
          cert: /etc/ssl/client.pem
          private_key: /etc/ssl/client-key.pem    # PKCS#8 only
          private_key_passphrase: "..."           # optional
          cert_reread_interval_seconds: 3600      # optional, for
  rotation
      headers:                          # optional, applied via
  setHeader
        X-Caller-Identity: "..."
  ```


README.md
```diff
@@ -264,9 +264,17 @@ nd_builtin_cache: true
 |-------|------|---------|-------------|
 | `url` | string | - | Base URL of the service |
 | `credentials.bearer.token` | string | - | Bearer token for authentication |
+| `credentials.client_tls.cert` | string | - | PEM file with the client certificate (mTLS) |
```
johanfylling: Presumably, the string value here is the path to the file? Something to clarify in the description?

johanfylling: We should also support java-native keystores/truststores (JKS).

johanfylling: And these should be possible to pass in via the SDK.

opa-services/build.gradle.kts
```diff
@@ -13,6 +13,8 @@ dependencies {
     implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml")
     implementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310")
     implementation("org.apache.commons:commons-compress:1.28.0")
+    implementation("org.bouncycastle:bcprov-jdk18on:1.84")
+    implementation("org.bouncycastle:bcpkix-jdk18on:1.84")
```
johanfylling: We need to do crypto that's not supported in the java standard libs? Preferably, we keep required dependencies on a downstream consumer's classpath to a minimum; if we can get away with dropping BC, we should.

johanfylling: Worst case, we could maybe load BC reflectively at runtime instead of an explicit dependency.

opa-services/README.md
```diff
@@ -91,6 +91,74 @@ status:
...
+        cert_reread_interval_seconds: 3600
+```
+
+| Field | Description |
+|-------|-------------|
+| `tls.ca_cert` | PEM file containing one or more trust roots for verifying the server. |
+| `tls.system_ca_required` | When `true`, the JVM's default trust store is also trusted in addition to `ca_cert`. |
+| `credentials.client_tls.cert` | PEM file containing the client certificate (and any intermediates). |
+| `credentials.client_tls.private_key` | PKCS#8 PEM file with the client private key. |
+| `credentials.client_tls.private_key_passphrase` | Passphrase for an encrypted PKCS#8 key. Omit for unencrypted keys. |
```
johanfylling: Can we pick up configuration keys from the environment, like in main OPA? User's might not want to add plaintext secrets to the config.

opa-services/README.md
```diff
@@ -91,6 +91,74 @@ status:
...
+| `credentials.client_tls.cert_reread_interval_seconds` | If set, the cert and key are reloaded from disk on this interval to support runtime rotation. |
+
+Only PKCS#8 PEM private keys are accepted. Convert PKCS#1 keys with:
```
johanfylling: This makes me think BC is *not* required.

opa-services/README.md
```diff
@@ -91,6 +91,74 @@ status:
...
+
+For keystores that cannot be expressed as files (HSM-backed keys, secret-manager-driven rotation, custom `KeyManager` chains), supply a fully constructed `SSLContext` directly. When set, file-based TLS fields are rejected during validation:
+
+```java
+SSLContext sslContext = buildSslContextFromHsm();
+
+Config.ServiceConfig service = new Config.ServiceConfig()
+    .setName("acmecorp")
+    .setUrl("https://policy.example.com")
+    .setSslContext(sslContext);
```
johanfylling, line 195: 👍

opa-services/src/main/java/io/github/open_policy_agent/opa/config/Config.java
```diff
@@ -463,12 +471,39 @@ public ServiceConfig setAllowInsecureTLS(boolean allowInsecureTLS) {
...
+    }
+
+    /**
+     * Programmatic override for the per-service {@link SSLContext}.
+     *
+     * <p>When set, it is used as-is and file-based TLS fields ({@link TlsConfig},
+     * {@link ClientTlsConfig}) are rejected during validation. Use this for keystores that can't be
+     * expressed in YAML (PKCS12 from an enterprise cert manager, HSM-backed keys, runtime rotation
+     * from an external secret manager, etc.).
+     */
```
johanfylling: I'd expect comments like these on the setters, as those are more user facing.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
```diff
@@ -49,10 +48,79 @@ public Set<String> validate(PluginManager manager) {
       if (credential != null && credential.getType().equals(Credential.Type.NOT_SUPPORTED)) {
         errors.add("Service '" + serviceName + "' has unsupported credential type");
       }
+
+      errors.addAll(validateTls(serviceName, service));
     }
     return errors;
   }
 
+  private Set<String> validateTls(String serviceName, Config.ServiceConfig service) {
```
johanfylling, line 58: Interesting choice of returning a set of error messages instead of throwing exceptions; which would be the more java-idiomatic style.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
```diff
@@ -49,10 +48,79 @@ public Set<String> validate(PluginManager manager) {
...
+    Set<String> errors = new HashSet<>();
+    java.util.function.Function<String, String> err =
```
johanfylling: [nit] Consider an import instead of using the fully qualified name gere.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
```diff
@@ -49,10 +48,79 @@ public Set<String> validate(PluginManager manager) {
...
+        msg -> "Service '" + serviceName + "' " + msg;
+
+    Config.ClientTlsConfig clientTls =
+        service.getCredentials() == null ? null : service.getCredentials().getClientTls();
+    boolean hasServerTls = SslContextBuilder.hasServerTls(service);
+    boolean hasClientTls = SslContextBuilder.hasClientTls(service);
+    boolean hasProgrammatic = service.getSslContext() != null;
+
+    if (service.isAllowInsecureTLS() && (hasServerTls || hasClientTls || hasProgrammatic)) {
+      errors.add(err.apply("sets allow_insecure_tls=true alongside other TLS config; remove one"));
```
johanfylling: Should we signal to configure what's desired rather than not configuring what's undesired by saying `choose one` instead.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
```diff
@@ -49,10 +48,79 @@ public Set<String> validate(PluginManager manager) {
...
+        errors.add(
+            err.apply(
+                "credentials.client_tls.cert_reread_interval_seconds requires both cert"
+                    + " and private_key"));
+      }
+    }
+
+    if (service.getCredentials() != null
+        && service.getCredentials().getBearer() != null
+        && hasClientTls) {
```
johanfylling: Interesting restriction; I don't recall bearer tokens and mTLS being mutually exclusive.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
```diff
@@ -61,63 +129,62 @@ public Plugin initialize(PluginManager manager) {
...
-          new Service(client, manager.getLogger())
-              .setName(service.getName())
-              .setUrl(service.getUrl())
-              .setCredentials(getCredential(service)));
+        plugin.services.put(
+            service.getName(),
+            new Service(client, manager.getLogger())
+                .setName(service.getName())
+                .setUrl(service.getUrl())
+                .setCredentials(getCredential(service)));
```
johanfylling: Looks like we're setting non-volatile fields on the service here. This is probably fine, as I'm guessing we're not concurrently updating services during runtime. But if services are assumed to be implicitly immutable, maybe we should formalize that and employ a builder pattern like we do for other things.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
```diff
@@ -61,63 +129,62 @@ public Plugin initialize(PluginManager manager) {
...
+      }
+    } catch (RuntimeException e) {
+      // Partial init failed: shut down the scheduler so any reload tasks already scheduled
+      // for earlier services don't outlive this initialize() call.
+      plugin.certReloadScheduler.shutdownNow();
```
johanfylling, line 229: No graceful shutdown? Maybe not required here, just checking.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
```diff
@@ -37,23 +43,96 @@ public abstract class BundleDownloader {
...
 
-  protected BundleDownloader(String name, PluginManager manager) {
+  /**
+   * Construct a BundleDownloader.
+   *
+   * @param name bundle name (used in log messages)
+   * @param manager the owning plugin manager
+   * @param httpClient the HTTP client to use for HTTP/HTTPS downloads; may be {@code null} if this
+   *     downloader only ever handles {@code file://} URIs or filesystem paths
+   * @param authService the {@link ServicePlugin.Service} providing credentials for HTTP downloads;
```
johanfylling: How come we only use the service for authentication? Could it not be used for all connectivity related functions and be the http-client holder? Or are there cases where we can have a client but no service?

opa-services/src/main/java/io/github/open_policy_agent/opa/tls/ReloadingX509KeyManager.java
```diff
@@ -0,0 +1,206 @@
...
+  private final AtomicReference<State> state = new AtomicReference<>();
+
+  /**
+   * Build and start a reloading key manager.
+   *
+   * @param certPath PEM cert-chain path
+   * @param keyPath PEM PKCS#8 private-key path
+   * @param keyPassphrase passphrase for encrypted keys (nullable)
+   * @param scheduler executor used to schedule periodic reloads
+   * @param rereadInterval how often to check the on-disk bytes
```
johanfylling: `rereadIntervalSeconds`?

opa-services/src/main/java/io/github/open_policy_agent/opa/tls/ReloadingX509KeyManager.java
```diff
@@ -0,0 +1,206 @@
...
+   * @param logger logger for reload events
+   * @param serviceName name of the owning service (for log messages)
+   */
+  public static ReloadingX509KeyManager create(
+      Path certPath,
+      Path keyPath,
+      char[] keyPassphrase,
+      ScheduledExecutorService scheduler,
+      long rereadIntervalSeconds,
```
johanfylling: Instead of primitive numbers with the unit relayed through argument naming and docs, consider using `java.time.Duration`, which is less prone to user errors.

opa-services/src/test/java/io/github/open_policy_agent/opa/config/ConfigTest.java
```diff
@@ -324,4 +324,64 @@ void config_toString_includesDiscovery() {
...
+    assertEquals("/etc/ssl/client-key.pem", clientTls.getPrivateKey());
+    assertEquals("secret", clientTls.getPrivateKeyPassphrase());
+    assertEquals(3600, clientTls.getCertRereadIntervalSeconds());
+  }
+
+  @Test
+  void config_mtlsDefaults() {
+    Config.TlsConfig tls = new Config.TlsConfig();
+    assertFalse(tls.isSystemCaRequired());
+    assertEquals(null, tls.getCaCert());
```
johanfylling: [nit] Could simply use `assertNull` here.

opa-services/src/main/java/io/github/open_policy_agent/opa/tls/ReloadingX509KeyManager.java
```diff
@@ -0,0 +1,206 @@
...
+      Logger logger,
+      String serviceName)
+      throws IOException, GeneralSecurityException {
+    ReloadingX509KeyManager mgr =
+        new ReloadingX509KeyManager(certPath, keyPath, keyPassphrase, logger, serviceName);
+    mgr.loadOrThrow();
+    scheduler.scheduleAtFixedRate(
```
johanfylling: We don't need to hold on to the returned future so we can shut it down properly (e.g. on shutdown or config reload, if that's supported)?

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/DiscoveryPlugin.java
```diff
@@ -62,24 +62,33 @@ public Set<String> validate(PluginManager manager) {
...
     Config.DiscoveryConfig discoveryConfig = manager.getConfig().getDiscovery();
     if (discoveryConfig != null) {
       String name = discoveryConfig.getName() != null ? discoveryConfig.getName() : "discovery";
+
+      ServicePlugin.Service svc = null;
+      HttpClient client = null;
+      Plugin raw = manager.getPlugin("services");
+      if (raw instanceof ServicePlugin) {
+        svc = ((ServicePlugin) raw).getService(discoveryConfig.getService());
+        if (svc != null) {
```
johanfylling: It's not an error to have a service name configured but then not finding the service?
Is there a risk of the `DiscoveryBundle` silently falling back on the default HTTP client without the user then realizing the configured TLS config isn't active?

opa-services/src/test/java/io/github/open_policy_agent/opa/plugins/ServicePluginTest.java
```diff
@@ -309,4 +309,200 @@ void validate_multipleServices_oneInvalid_returnsError() {
...
+        new PluginManager.Builder()
+            .withId("t")
+            .withStore(store)
+            .withConfig(config)
+            .withLogger(mockLogger)
+            .build();
+
+    Set<String> errors = new ServicePlugin().validate(manager);
+    assertTrue(errors.stream().anyMatch(e -> e.contains("cert_reread_interval_seconds must be >= 0")));
+  }
```
johanfylling: [nit] There's a bunch of near-identical tests here. Maybe we can consolidate them into one parameterized test?

johanfylling (approved):
Thank you!

Mostly comments and nits; but it would be nice to not pull in the bouncycastle dependency.

README.md
sspaink: Added clarifying comment 👍

sspaink: Added tls.truststore.{path,password,type} and credentials.client_tls.keystore.{path,password,key_password,type}. Type defaults to PKCS12, or JKS if the path ends in .jks.

opa-services/build.gradle.kts
sspaink: Dropped bouncycastle

opa-services/README.md
sspaink: Good idea, ${VAR} is now supported

opa-services/build.gradle.kts
sspaink: Dropped bouncycastle

README.md
sspaink: Good idea, docs updated

sspaink: Done. Added tls.truststore.{path,password,type} and credentials.client_tls.keystore.{path,password,key_password,type}. Type defaults to PKCS12, or JKS if the path ends in .jks.

opa-services/src/test/java/io/github/open_policy_agent/opa/plugins/ServicePluginTest.java
sspaink: Consolidated 👍

opa-services/src/main/java/io/github/open_policy_agent/opa/config/Config.java
sspaink: Moved 👍

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
sspaink: Followed existing pattern in other plugins, could refactor later

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
sspaink: Check removed

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/ServicePlugin.java
sspaink: ServicePlugin.stop() already drains the cert-reload scheduler

opa-services/src/main/java/io/github/open_policy_agent/opa/tls/ReloadingX509KeyManager.java
sspaink: Done. ReloadingX509KeyManager.create(...) and SslContextBuilder both take a Duration now

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
sspaink: refactored. ServicePlugin.Service now owns the HttpClient

PR #71, opened 2026-06-23
fix: harden bundle downloader against security vulnerabilities

### LINKED ISSUE #67: Bundle download: size cap and content-type check
`BundleDownloader.handleHttpDownload` calls `BodyHandlers.ofByteArray()` which buffers the entire response into memory before any validation, and then hands the bytes to `TarballBundleLoader` which decompresses without a ratio check. A hostile or misconfigured server can serve an arbitrarily large body or a decompression bomb and OOM the process.

* Add a configurable max bundle size; reject responses whose `Content-Length` exceeds it before reading, and stop reading streams that exceed it mid-flight.
* Verify response `Content-Type` matches `application/vnd.openpolicyagent.bundles` (or a configured allowlist) before treating the body as a bundle.
* Cap the decompressed size in `TarballBundleLoader` to guard against zip-bomb-style tar.gz inputs.
* Tests: oversized response rejected, wrong content-type rejected, decompression bomb rejected.

polachandu:
Closes #67

- Validated Content-Type header on HTTP bundle responses; reject anything that is not application/vnd.openpolicyagent.bundles
- Added configurable max_size_bytes limit (default 512 MB) to BundleConfig; introduced upfront via Content-Length header and mid-stream via a custom BodySubscriber to prevent memory exhaustion from oversized responses
- Added decompressed size cap in TarballBundleLoader tracked cumulatively across all tar entries to prevent decompression bomb attacks
- Added tests covering wrong/missing content-type, Content-Length limit, mid-stream size limit, decompression bomb, and cumulative entry limits

### DIFF
--- a/opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
+++ b/opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
@@ -11,20 +11,33 @@ import java.nio.file.Path;
 import java.util.zip.GZIPInputStream;
 
 public class TarballBundleLoader implements BundleLoader {
+  private static final long DEFAULT_MAX_DECOMPRESSED_SIZE = 512 * 1024 * 1024L; // 512 MB
+
   private final Path path;
   private final byte[] data;
   private final String id;
+  private final long maxDecompressedBytes;
 
   public TarballBundleLoader(String id, Path path) {
+    this(id, path, DEFAULT_MAX_DECOMPRESSED_SIZE);
+  }
+
+  public TarballBundleLoader(String id, Path path, long maxDecompressedBytes) {
     this.path = path;
     this.data = null;
     this.id = id;
+    this.maxDecompressedBytes = maxDecompressedBytes;
   }
 
   public TarballBundleLoader(String id, byte[] data) {
+    this(id, data, DEFAULT_MAX_DECOMPRESSED_SIZE);
+  }
+
+  public TarballBundleLoader(String id, byte[] data, long maxDecompressedBytes) {
     this.path = null;
     this.data = data;
     this.id = id;
+    this.maxDecompressedBytes = maxDecompressedBytes;
   }
 
   public Bundle load(Store store) {
@@ -59,26 +72,31 @@ public class TarballBundleLoader implements BundleLoader {
     try (GZIPInputStream gzipIn = new GZIPInputStream(in);
         TarArchiveInputStream tarIn = new TarArchiveInputStream(gzipIn, true)) {
       org.apache.commons.compress.archivers.tar.TarArchiveEntry entry;
+      long totalDecompressedBytes = 0;
       while ((entry = tarIn.getNextTarEntry()) != null) {
         String entryName =
             entry.getName().startsWith("/") ? entry.getName().substring(1) : entry.getName();
         if (!entry.isDirectory() && entryName.equals("plan.json")) {
-          byte[] entryBytes = tarIn.readAllBytes();
+          byte[] entryBytes = readWithLimit(tarIn, maxDecompressedBytes - totalDecompressedBytes);
+          totalDecompressedBytes += entryBytes.length;
           assembler.loadPlan(new ByteArrayInputStream(entryBytes));
         }
         if (!entry.isDirectory()
             && (entryName.equals("data.json") || entryName.endsWith("/data.json"))) {
-          byte[] entryBytes = tarIn.readAllBytes();
+          byte[] entryBytes = readWithLimit(tarIn, maxDecompressedBytes - totalDecompressedBytes);
+          totalDecompressedBytes += entryBytes.length;
           int lastSlash = entryName.lastIndexOf('/');
           String dataPath = lastSlash < 0 ? "" : entryName.substring(0, lastSlash);
           assembler.loadData(dataPath, new ByteArrayInputStream(entryBytes));
         }
         if (!entry.isDirectory() && entryName.equals(".manifest")) {
-          byte[] entryBytes = tarIn.readAllBytes();
+          byte[] entryBytes = readWithLimit(tarIn, maxDecompressedBytes - totalDecompressedBytes);
+          totalDecompressedBytes += entryBytes.length;
           assembler.loadManifest(new ByteArrayInputStream(entryBytes));
         }
         if (!entry.isDirectory() && entryName.endsWith(".rego")) {
-          byte[] entryBytes = tarIn.readAllBytes();
+          byte[] entryBytes = readWithLimit(tarIn, maxDecompressedBytes - totalDecompressedBytes);
+          totalDecompressedBytes += entryBytes.length;
           assembler.addRego(entryName, new String(entryBytes));
         }
       }
@@ -86,5 +104,18 @@ public class TarballBundleLoader implements BundleLoader {
     } catch (IOException e) {
       throw new IllegalArgumentException("Error extracting bundle: " + e.getMessage(), e);
     }
+  }
+
+  private static byte[] readWithLimit(InputStream in, long remaining) throws IOException {
+    if (remaining <= 0) {
+      throw new IOException(
+          "Decompressed bundle size exceeds the configured limit");
+    }
+    byte[] bytes = in.readNBytes((int) Math.min(remaining + 1, Integer.MAX_VALUE));
+    if (bytes.length > remaining) {
+      throw new IOException(
+          "Decompressed bundle size exceeds the configured limit");
+    }
+    return bytes;
   }
 }
--- a/opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
+++ b/opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
@@ -1,5 +1,6 @@
 package io.github.open_policy_agent.opa.bundle;
 
+import io.github.open_policy_agent.opa.config.Config;
 import io.github.open_policy_agent.opa.storage.Store;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 
@@ -11,7 +12,8 @@ import java.nio.file.Path;
 import java.util.zip.GZIPInputStream;
 
 public class TarballBundleLoader implements BundleLoader {
-  private static final long DEFAULT_MAX_DECOMPRESSED_SIZE = 512 * 1024 * 1024L; // 512 MB
+  private static final long DEFAULT_MAX_DECOMPRESSED_SIZE =
+      Config.BundleConfig.DEFAULT_MAX_SIZE_BYTES;
 
   private final Path path;
   private final byte[] data;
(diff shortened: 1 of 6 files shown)


opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
```diff
@@ -48,6 +53,22 @@ public abstract class BundleDownloader {
...
   protected long lastModifiedTime = 0;
+  protected long maxSizeBytes = Config.BundleConfig.DEFAULT_MAX_SIZE_BYTES;
+
+  // Content types accepted for bundle responses. The OPA server sends
+  // "application/vnd.openpolicyagent.bundles", but bundles are very commonly hosted on static
+  // object storage (S3, GCS, nginx) that serves tarballs as gzip/octet-stream, so those are
+  // accepted too. An absent/blank Content-Type is also tolerated. The point of the check is to
+  // reject obviously-wrong responses (e.g. an HTML error page or login redirect) before they are
+  // parsed as a bundle.
+  private static final Set<String> ALLOWED_CONTENT_TYPES =
```
polachandu, line 60: This is a set where the content types are accepted. Comment here if I have to add anything or change the logic of it.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
```diff
@@ -329,6 +364,88 @@ private void handleHttpDownload(URI uri) throws IOException, InterruptedExceptio
...
+  }
+
+  /**
+   * Returns a BodyHandler that accumulates response bytes into a byte array, throwing
+   * BundleSizeLimitException mid-stream if the total exceeds maxBytes.
+   */
+  private static HttpResponse.BodyHandler<byte[]> sizeLimitedBodyHandler(long maxBytes) {
+    return responseInfo -> {
+      SizeLimitedSubscriber subscriber = new SizeLimitedSubscriber(maxBytes);
+      return HttpResponse.BodySubscribers.fromSubscriber(subscriber, SizeLimitedSubscriber::getResult);
```
sspaink: **Blocking: mid-stream cancel hangs the poller forever on a large response.**

`fromSubscriber(subscriber, getResult)` only completes the response-body future when the subscriber emits a terminal signal (`onComplete`/`onError`). When `onNext` hits the limit it calls `subscription.cancel()` (line 413) and stashes the error — but cancelling a streamed body produces *no* terminal signal, so `getResult` is never applied and the body future never completes. `httpClient.send` then blocks the (single-threaded) poll scheduler indefinitely.

I verified this: the existing `download_bodyExceedsLimitMidStream_fails` test only passes because its body is tiny and arrives in one chunk, so `onComplete` races in alongside the cancel. Streaming ~128 MB chunked against a 1 MB limit makes `getInitialActivation().get(15s)` throw `TimeoutException` — the download never returns. So against a genuinely oversized response (exactly the DoS this PR targets) the downloader hangs and holds the buffered bytes instead of failing fast.

Fix: implement `HttpResponse.BodySubscriber<byte[]>` directly and complete its `getBody()` stage exceptionally on overflow, rather than `fromSubscriber` + `cancel()`. Please also add a test that streams a multi-chunk body well over the limit and asserts the future completes promptly.

opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
```diff
@@ -59,26 +74,31 @@ private Bundle createBundleFromStream(String id, InputStream in, Store store) {
     try (GZIPInputStream gzipIn = new GZIPInputStream(in);
         TarArchiveInputStream tarIn = new TarArchiveInputStream(gzipIn, true)) {
       org.apache.commons.compress.archivers.tar.TarArchiveEntry entry;
+      long totalDecompressedBytes = 0;
```
sspaink, line 77: **Cumulative cap only counts matched entries.**

`readWithLimit` / `totalDecompressedBytes` are only invoked inside the `if` blocks for `plan.json`, `data.json`, `.manifest`, and `*.rego`. Any other entry name is never read and never counted, so a tarball padded with one enormous non-policy entry sails past this budget. `getNextTarEntry()` still decompresses the skipped bytes, so memory stays bounded but CPU/time does not — a decompression-bomb CPU DoS remains possible.

The PR description says size is "tracked cumulatively across all tar entries," which overstates what this does. Consider counting every entry's decompressed size (or bytes read from the tar stream) regardless of whether the entry is one we load.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
```diff
@@ -304,7 +330,7 @@ private void handleHttpDownload(URI uri) throws IOException, InterruptedExceptio
 
     HttpRequest request = requestBuilder.build();
     HttpResponse<byte[]> response =
-        httpClient.send(request, HttpResponse.BodyHandlers.ofByteArray());
+        httpClient.send(request, sizeLimitedBodyHandler(maxSizeBytes));
```
sspaink: **No upfront Content-Length check (contrary to PR description).**

The description says oversized responses are rejected "upfront via Content-Length header," but there's no Content-Length inspection — enforcement is entirely mid-stream, the path broken in the other comment. A cheap early reject (`Content-Length > maxSizeBytes` → fail) would both match the description and avoid buffering for honest servers that advertise their size.

opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
```diff
@@ -87,4 +107,17 @@ private Bundle createBundleFromStream(String id, InputStream in, Store store) {
       throw new IllegalArgumentException("Error extracting bundle: " + e.getMessage(), e);
     }
   }
+
+  private static byte[] readWithLimit(InputStream in, long remaining) throws IOException {
```
sspaink, line 114: _Minor / edge:_ `Math.min(remaining + 1, Integer.MAX_VALUE)` overflows to a negative value if `maxSizeBytes` is configured near `Long.MAX_VALUE`, making `readNBytes` throw and breaking legitimate loads. Separately, `buffer.size()` (an `int`) plus `ByteArrayOutputStream`'s ~2 GB ceiling mean the mid-stream cap can't actually be set above 2 GB. Both are out-of-range edge cases, not default-config bugs — worth a guard or a documented bound on `max_size_bytes`.

sspaink (changes requested):
Thanks for hardening the bundle downloader — the Content-Type check and the per-entry decompression cap are solid, and the test suite passes. One blocking issue though: the mid-stream size limit hangs the poller on large responses (verified empirically), which defeats the DoS protection this PR is meant to add. Details inline.

opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
polachandu: That's a good catch! Fixing it.

opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
polachandu: Great to see finding the OOR edge cases. Fixed it.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
```diff
@@ -303,29 +320,101 @@ private void handleHttpDownload(URI uri) throws IOException, InterruptedExceptio
...
+                if (!isAcceptableContentType(contentType)) {
+                  String errorMsg = "Unexpected Content-Type: '" + contentType + "'";
+                  manager.getLogger().error("Bundle '%s': %s", name, errorMsg);
+                  if (!initialActivation.isDone()) {
+                    initialActivation.completeExceptionally(new RuntimeException(errorMsg));
+                  }
+                  return;
+                }
+                response.headers().firstValue("ETag").ifPresent(newEtag -> this.etag = newEtag);
+                activateBundle(response.body());
```
sspaink: **Blocking: activation failure leaves `initialActivation` pending forever.**

`activateBundle` re-throws a `RuntimeException` on any load failure (see `BundlePlugin.activateBundle`). Here it runs inside the `whenComplete` action, *before* `initialActivation.complete(null)`. When it throws, the exception just completes the discarded future returned by `whenComplete` — `initialActivation` is never touched, so `getInitialActivation().get()` blocks indefinitely instead of failing.

The old synchronous code propagated the throw up to `downloadBundle`'s catch, which called `initialActivation.completeExceptionally(e)`. That path is lost in the async rewrite.

Failure scenario: server returns 200 with a valid `Content-Type` but a malformed tarball — or a decompression bomb that trips the new `TarballBundleLoader` size cap (exactly this PR's target). Activation throws, and a caller awaiting initial activation hangs. None of the current HTTP tests cover this (they all activate successfully).

Fix: wrap `activateBundle(...)` in a try/catch (or complete `initialActivation` exceptionally in the `throwable`/catch path) so an activation error surfaces to the future. A test that serves a 200 with a body that fails to load and asserts the future completes exceptionally would lock this down.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
```diff
@@ -303,29 +320,101 @@ private void handleHttpDownload(URI uri) throws IOException, InterruptedExceptio
...
+              throw new BundleSizeLimitException(
+                  "Content-Length "
+                      + contentLength.getAsLong()
+                      + " exceeds limit of "
+                      + cappedMax
+                      + " bytes");
+            });
       }
+      return HttpResponse.BodySubscribers.mapping(
+          HttpResponse.BodySubscribers.ofByteArray(),
```
sspaink: **Body-size limit is enforced only after the full body is buffered — no genuine mid-stream cap.**

`BodySubscribers.mapping(ofByteArray(), fn)` applies the size check `fn` only *after* `ofByteArray()` has accumulated the entire body into memory; there's no cancellation as bytes arrive. So the `bytes.length > cappedMax` guard rejects *after the fact* rather than mid-stream.

The upfront `Content-Length` check helps for honest servers, but a malicious server can simply omit `Content-Length` (or use chunked encoding) and stream a body far larger than `max_size_bytes`. `ofByteArray` will buffer it all — up to the ~2 GB array ceiling — before the mapping throws, which undercuts the memory-exhaustion protection this PR is meant to provide.

Fix: implement `HttpResponse.BodySubscriber<byte[]>` directly, track cumulative bytes in `onNext`, and complete the body stage exceptionally (cancelling the subscription) the moment the running total exceeds `cappedMax`. A test that streams a multi-chunk body over the limit *with no `Content-Length`* and asserts prompt failure would cover it. (Note: the current `download_...MidStream` tests only send ~1.28 MB, so they never exercise true buffering pressure.)

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
polachandu: Applying the fix.

sspaink (approved):
Thank you!

PR #93, opened 2026-06-24
Migrate to jackson 3

sspaink:
Migrate the SDK from Jackson 2.x (`com.fasterxml.jackson.*`) to Jackson 3.x (`tools.jackson.*`) so we can pick up `com.networknt:json-schema-validator:3.0.5`, which dropped its Jackson 2 line. https://github.com/open-policy-agent/java-opa-sdk/pull/52

Swap the abandoned `com.github.java-json-tools:json-patch` for an inline RFC 6902 implementation in `JsonBuiltins` (no new dependency).


sspaink:
Let's revisit this in the future, not required atm

PR #103, opened 2026-07-06
ci: add dependency review to PRs

sspaink:
## What

Adds supply-chain dependency scanning to CI:

- **`dependency-review` job** in the PR check using `actions/dependency-review-action`, gated on `high`+ severity advisories, posting a summary comment on failure. Wired into the existing `pr-check-summary` gate.
- **`dependency-submission.yml`** workflow that submits the resolved Gradle dependency graph to GitHub's Dependency Graph on pushes to `main`. This gives the PR review a baseline to diff against and enables Dependabot alerts for transitive dependencies.

## Why

We publish artifacts to Maven Central, so newly introduced vulnerable/disallowed dependencies should be caught at PR time.

## Notes

- Actions are hash-pinned to match the repo's zizmor policy.
- The dependency-review baseline lags one merge: on the first PR after this lands, the base graph won't exist until `dependency-submission.yml` has run once on `main`. It self-corrects after that.
- Requires *Settings -> Code security -> Dependency graph* enabled (default for public repos).

### DIFF
--- a/.github/workflows/dependency-submission.yml
+++ b/.github/workflows/dependency-submission.yml
@@ -1,0 +1,29 @@
+name: Dependency Submission
+
+# Submits the resolved Gradle dependency graph to GitHub's Dependency Graph.
+# This populates the baseline for `dependency-review` on pull requests and
+# enables Dependabot alerts for transitive dependencies.
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


kroekle (approved):
LGTM

PR #107, opened 2026-07-06
ci: test and type-check Rego policy on PRs

sspaink:
## What

Adds a `rego-policy` job to the PR check that type-checks and tests the Rego under `tools/policy/`, wired into the `pr-check-summary` gate:

- `opa check --strict tools/policy/` — catches parse/type errors.
- `opa test tools/policy/ -v` — runs the existing `pr_check_test.rego` suite (9 tests).

## Why

`tools/policy/pr-check/pr_check.rego` is the policy that drives this workflow's **own** change filter (deciding which module test jobs run). It had a test file but nothing in CI executed it, so a broken rule or a syntax/type error there could silently break CI gating and merge unnoticed.

## Notes

- Scoped to `tools/policy/` only. The other `.rego` files in the repo are Java test fixtures/examples and are intentionally left out of these checks.
- Uses the `setup-opa` action already used elsewhere in this workflow (same hash pin).
- `opa fmt` was intentionally **not** added as a gate: formatting rules drift across OPA releases and the workflow pulls `version: latest`, which would make a fmt check flaky. `check` + `test` are deterministic.
- Verified locally: `opa test` 9/9 pass, `opa check --strict` clean.


kroekle (approved):
LGTM

PR #110, opened 2026-07-08
feat(opa-proto): support proto-format plan bundles (opa build --format=proto)

### LINKED ISSUE #101: Support proto-format plan bundles (opa build --format=proto)
## Summary

Add support for reading **protobuf-format plan bundles** produced by OPA's new `opa build --format=proto` flag ([open-policy-agent/opa#8825](https://github.com/open-policy-agent/opa/pull/8825), merged).

## Background

OPA now emits IR plan bundles in two wire formats:

| `--format` | Plan file | Manifest file |
|------------|-----------|---------------|
| `json` (default) | `/plan.json` | `/.manifest` |
| `proto` (new) | `/plan.pb` | `/.manifest.pb` |

The protobuf encoding is defined by the schemas added in [open-policy-agent/opa#8775](https://github.com/open-policy-agent/opa/pull/8775):

- `v1/ir/plan.proto` — the IR plan
- `v1/bundle/manifest.proto` — the bundle manifest

The proto form is significantly smaller on the wire (per the PR, a minimal policy produces `plan.pb` 469 bytes vs `plan.json` 2098 bytes; `.manifest.pb` 21 bytes vs `.manifest` 59 bytes). OPA's own bundle reader auto-detects either form and rejects mixed-format bundles.

## Current state in this SDK

The SDK today only understands the JSON forms:

- `FileSystemBundleLoader` hard-codes the filenames `plan.json` and `.manifest`.
- The `PolicyReader` SPI (implemented by `JacksonPolicyReader` in `opa-jackson`) parses the JSON IR plan.
- The `BundleParser` SPI (implemented by `JacksonBundleParser`) parses the JSON `.manifest`.

There is no path to load a bundle built with `--format=proto`.

## Proposed work

1. **Bundle detection** — teach `FileSystemBundleLoader` (and any other `BundleLoader` implementations) to detect `plan.pb` / `.manifest.pb` in addition to `plan.json` / `.manifest`, matching OPA's auto-detection and mixed-format rejection semantics.
2. **Proto plan decoding** — add a `PolicyReader` implementation that decodes `plan.pb` per `v1/ir/plan.proto` into the existing `Policy` model.
3. **Proto manifest decoding** — add a `BundleParser` (or equivalent) capable of decoding `.manifest.pb` per `v1/bundle/manifest.proto`.
4. **Schema source & codegen** — decide how to vendor/generate the Java bindings from the `.proto` schemas (e.g. `protobuf-gradle-plugin`), pinning the schema version to a specific OPA release for reproducible output. Keep protobuf as an optional dependency so the core `opa-evaluator` stays lightweight — likely a new module (e.g. `opa-proto`) paralleling `opa-jackson`.
5. **Tests** — round-trip fixtures built with both `--format=json` and `--format=proto` from the same policy, asserting identical evaluation results; plus a mixed-format-bundle rejection test.

## Open questions

- Should proto support live in its own module (`opa-proto`) or be folded into an existing one?
- Which OPA release should we pin the `.proto` schemas to?
- Do we need write support (emitting `.pb` bundles), or is read-only sufficient for the SDK's use cases?

## References

- OPA PR adding the flag: https://github.com/open-policy-agent/opa/pull/8825
- OPA PR adding the proto schemas: https://github.com/open-policy-agent/opa/pull/8775

sspaink:
Adds support for reading **protobuf-format plan bundles** produced by `opa build --format=proto` (`plan.pb` / `.manifest.pb`), decoding them into the existing IR model so they evaluate identically to their JSON counterparts.

Proto bundles keep `data.json` as JSON, so proto support is an **optional add-on** alongside a JSON provider (e.g. opa-jackson), not a replacement.

Closes #101

`opa build --format=proto` isn't in a tagged OPA release yet, so `go.mod` is pinned to a **temporary commit** (`f6092b9ce`, ~`v1.19.0-dev`; PRs open-policy-agent/opa#8825 and #8775).


opa-evaluator/src/main/java/io/github/open_policy_agent/opa/bundle/BundleAssembler.java
```diff
@@ -132,6 +154,63 @@ public void loadManifest(InputStream in) throws IOException {
...
+   * @param manifestJson source for {@code .manifest}, or {@code null} if absent
+   * @param manifestProto source for {@code .manifest.pb}, or {@code null} if absent
+   * @throws IllegalArgumentException if the bundle mixes wire formats
+   * @throws IOException if a source cannot be opened or its contents cannot be parsed
+   */
+  public void loadPlanAndManifest(
+      InputStreamSource planJson,
+      InputStreamSource planProto,
+      InputStreamSource manifestJson,
+      InputStreamSource manifestProto)
```
johanfylling, line 178: We could consider making this a bit less rigid, and take a vararg of some type wrapping a `InputStreamSource` together with type and format meta.

This is probably fine though as it's mostly an internal API and as we don't have any additional formats planned; but then again, we also didn't have protobuf on the backlog not too long ago ;)

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/bundle/BundleAssembler.java
```diff
@@ -38,39 +36,53 @@ public class BundleAssembler {
...
 
+  /**
+   * Holds the optional proto decoder, resolved lazily on first proto-format load.
+   *
+   * <p>Deferring resolution (rather than a {@code static final} field on {@link BundleAssembler})
+   * keeps a proto-specific misconfiguration — no {@code opa-proto} module, two providers, or a
+   * provider that fails to instantiate — from running in {@code BundleAssembler}'s own class
+   * initializer, where it would break <em>all</em> bundle loading, including JSON-only bundles that
+   * never touch proto. The holder class initializes only when {@link #requireProtoDecoder} first
+   * references it, so any such failure surfaces only when a proto bundle is actually read.
```
johanfylling: Consider keeping these overly verbose defensive comments to a minimum.  They bury the lead somewhat, and it's often easier to understand the context from just reading the code. E.g. Here, I think it's enough to simply say: `Deferring decoder resulution to time of bundle load.`

johanfylling: Another though: do we really need to defer this error? Simply calling `loadOptional` will result in either a decoder instance or `null` when the class-path is well configured. It only throws an error if there are more than one implementations on the classpath, in which case we should want to fail early, just as for `POLICY_READER` and `BUNDLE_PARSER`, right? A misconfigured classpath is a initialization issue that is unlikely to resolve itself later in the app's life-cycle.

opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
```diff
@@ -71,29 +71,43 @@ public Bundle load(Store store) {
...
+    byte[] planJson = null;
+    byte[] planProto = null;
+    byte[] manifestJson = null;
+    byte[] manifestProto = null;
     try (GZIPInputStream gzipIn = new GZIPInputStream(in);
         TarArchiveInputStream tarIn = new TarArchiveInputStream(gzipIn, true)) {
       org.apache.commons.compress.archivers.tar.TarArchiveEntry entry;
       long totalDecompressedBytes = 0;
       while ((entry = tarIn.getNextTarEntry()) != null) {
+        if (entry.isDirectory()) {
```
johanfylling, line 86: Won't this make the `if (!entry.isDirectory())` check further down redundant?

opa-proto/src/test/java/io/github/open_policy_agent/opa/proto/StructConverterTest.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Verifies that protobuf Struct/Value conversion matches the Java types the JSON {@code .manifest}
+ * reader produces — in particular that integral numbers become Integer/Long (not Double), so
+ * metadata is identical across the two bundle wire formats.
+ */
+class StructConverterTest {
```
johanfylling, line 23: Maybe not enough test cases to bother, but some of these are very similar, and could be consolidated into parameterized tests.

opa-proto/src/test/java/io/github/open_policy_agent/opa/proto/ProtoBundleParityTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Objects;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Round-trip parity: the {@code authz-proto} and {@code authz-json} bundles are built from the same
+ * policy and data at the pinned OPA commit, differing only in wire format. Loading and evaluating
+ * either must produce identical results, proving the proto decoder is faithful to OPA's output.
+ */
+class ProtoBundleParityTest {
```
johanfylling, line 35: Consider parameterized tests.

opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
```diff
@@ -71,29 +71,43 @@ public Bundle load(Store store) {
...
-        } else if (!entry.isDirectory()
-            && (entryName.equals("data.json") || entryName.endsWith("/data.json"))) {
+        if (entryName.equals(BundleFormat.PLAN_JSON)) {
+          planJson = readWithLimit(tarIn, maxDecompressedBytes - totalDecompressedBytes);
+          totalDecompressedBytes += planJson.length;
+        } else if (entryName.equals(BundleFormat.PLAN_PROTO)) {
+          planProto = readWithLimit(tarIn, maxDecompressedBytes - totalDecompressedBytes);
+          totalDecompressedBytes += planProto.length;
+        } else if (entryName.equals(BundleFormat.MANIFEST_PROTO)) {
+          // Check .manifest.pb before .manifest so the longer suffix wins.
```
johanfylling: We're doing exact string matching, isn't order irrelevant?

.github/workflows/pull-request.yml
```diff
@@ -248,6 +248,36 @@ jobs:
...
+      - uses: actions/setup-java@1bcf9fb12cf4aa7d266a90ae39939e61372fe520 # v5.4.0
+        with:
+          distribution: temurin
+          java-version: 21
+      - uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5.6.0
+        with:
+          go-version-file: tools/generate-compliance-tests/go.mod
+      # The vendored .proto files in opa-proto must always match the OPA version
+      # pinned in tools/generate-compliance-tests/go.mod (the source of truth).
```
johanfylling, line 329: Are we generating and running protobuf variants of all the compliance test cases too?

cli/build.gradle.kts
```diff
@@ -15,6 +15,9 @@ dependencies {
 
     runtimeOnly(project(":opa-builtins"))
     runtimeOnly(project(":opa-jackson"))
+    // Enables the CLI to read proto-format bundles (opa build --format=proto). opa-jackson still
+    // handles data.json and JSON-format plans/manifests; opa-proto decodes plan.pb/.manifest.pb.
```
johanfylling: We don't provide rationales for the other dependencies; not sure if this is necessary.

opa-proto/src/main/proto/README.md
```diff
@@ -0,0 +1,55 @@
...
+[`tools/generate-compliance-tests/go.mod`](../../../../tools/generate-compliance-tests/go.mod),
+which is the single place this SDK declares which OPA revision it targets.
+
+At the time of writing that is a temporary commit pin
+(`v1.18.2-0.20260706175033-f6092b9ce472`, i.e. commit `f6092b9ce` — the merge of
+[open-policy-agent/opa#8825](https://github.com/open-policy-agent/opa/pull/8825),
+schemas from [#8775](https://github.com/open-policy-agent/opa/pull/8775)), because
+`opa build --format=proto` has not yet shipped in a tagged release. Once a release
+containing the flag is available (targeting v1.19.0), bump `go.mod` to that release
+and re-vendor — see below.
```
johanfylling: Protobuf n00b question: how backwards-compatible are these schemas? Will a newer java-opa-sdk always be compatible with an older OPA bundle, or is some lock-stepping required for the opa-proto lib on the app's classpath?

opa-proto/src/main/proto/v1/ir/plan.proto
```diff
@@ -0,0 +1,364 @@
...
+// On the Go side, `ir.Location` is embedded into every concrete Stmt
+// implementation, so `encoding/json` flattens File/Col/Row into the
+// emitted JSON body. The proto promotes those fields to the envelope
+// because that's both more idiomatic protobuf and lets every body
+// message start its own field numbering at 1.
+//
+// Case-number assignments (4–37) are a stability commitment. Field
+// numbers 1–3 are reserved for the location triple. New cases must be
+// added with the next unused number; existing numbers must never be
+// repurposed.
```
johanfylling, line 114: Ok, maybe this answers my question on README.md somewhat; there's not much difference to how version drift is handled for json plans: as long as the protobuf plan doesn't contain new, unknown stuff, we're pretty much version agnostic.

.github/workflows/pull-request.yml
johanfylling: Looks like no. This is left for a future task?

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/bundle/ProtoBundleDecoder.java
```diff
@@ -0,0 +1,54 @@
...
+ * {@code opa build --format=proto}.
+ *
+ * <p>OPA emits IR plan bundles in two wire formats. The default JSON form stores the plan at
+ * {@code /plan.json} and the manifest at {@code /.manifest}; the proto form stores them at
+ * {@code /plan.pb} and {@code /.manifest.pb} respectively. This SPI decodes the proto form into the
+ * same in-memory model the JSON {@link io.github.open_policy_agent.opa.ir.PolicyReader} produces, so
+ * evaluation is identical regardless of the on-disk format.
+ *
+ * <p>Unlike {@link io.github.open_policy_agent.opa.ir.PolicyReader} and {@link BundleParser} — of
+ * which exactly one implementation must be registered — this decoder is <em>optional</em>. The core
```
johanfylling, line 19: If there is a `ProtoBundleDecoder` on the classpath, wouldn't that then allow for at least `PolicyReader` to also be optional?

But seeing as json is is the default and still required for data, it's not the whole world to keep this stricter requirement.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/mapper/AnnotationIntrospectors.java
```diff
@@ -21,26 +20,7 @@ static AnnotationIntrospector get() {
...
-          }
-        names.append(impls.get(i).getClass().getName());
-      }
-      throw new IllegalStateException(
-          "Multiple AnnotationIntrospector implementations found on the classpath: "
-              + names
-              + ". Only one provider may be registered.");
-    }
-    return impls.get(0);
+    return Services.loadAtMostOne(AnnotationIntrospector.class)
```
johanfylling, line 23: Not related to this PR, but I wonder if we could simplify the `AnnotationIntrospector` interface? E.g. instead of asking it a bunch of questions, we simply give it a `Class` and it hands back a `ClassInfo`.

johanfylling (approved):
Thanks!

opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
sspaink: your right, misleading comment removed

opa-services/src/main/java/io/github/open_policy_agent/opa/bundle/TarballBundleLoader.java
sspaink: good catch, removed the redundant else check

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/bundle/BundleAssembler.java
sspaink: Good idea, updated to follow the same pattern as `POLICY_READER` and `BUNDLE_PARSER`

opa-proto/src/main/proto/v1/ir/plan.proto
sspaink: Exactly, same story as json

.github/workflows/pull-request.yml
sspaink: Your right there isn't a protobuf variant, but there should be! I hadn't thought about it. This `verify-proto-vendor` job here just checks the vendored proto schemas match the go.mod pinned OPA version. Probably better if it happens in this PR, will work on adding a proto variant.

cli/build.gradle.kts
sspaink: removed 👍

.github/workflows/pull-request.yml
sspaink: protobuf variant test added! Does have the negative side effect of having to regenerate all of the compliance tests to match the latest OPA so the file diff is crazy.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/mapper/AnnotationIntrospectors.java
sspaink: follow up issue created: https://github.com/open-policy-agent/java-opa-sdk/issues/178

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/bundle/BundleAssembler.java
sspaink: Lets just keep it like this for now, easily refactor it if a third option ever shows up

johanfylling (approved):
LGTM 👍

PR #111, opened 2026-07-09
Auto-discover plugins in status report

### LINKED ISSUE #82: Auto-discover plugins in status report
`StatusPlugin.Status.buildStatusReport` lists plugins by hardcoded name (`bundles`, `decision_logs`, `status`, `discovery`). Any plugin registered through the SDK's plugin extension points is silently dropped from the status payload — making the status report incomplete for non-trivial deployments.

* Iterate over all plugins registered with `PluginManager` (add an accessor if needed) instead of probing a fixed list of names.
* Include the SDK's built-ins by default; custom plugins automatically appear once registered.
* Tests covering the built-in set and a custom plugin showing up in the report.

Develop-KIM:
Fixes #82

### Summary
- Add a `PluginManager` accessor for the currently registered plugin names.
- Build status report plugin statuses by iterating registered plugins instead of a hardcoded built-in list.
- Cover custom registered plugins appearing in the status report.

### Verification
- `./gradlew :opa-services:test`
- `./gradlew :opa-services:check`
- `git diff --check`

### Disclosure
This contribution was prepared with AI assistance and reviewed before submission.

### DIFF
--- a/opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/StatusPlugin.java
+++ b/opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/StatusPlugin.java
@@ -185,10 +185,9 @@ public final class StatusPlugin implements Plugin {
 
       // Add plugin statuses (only for registered plugins)
       ObjectNode plugins = MAPPER.createObjectNode();
-      addPluginStatusIfRegistered(plugins, "bundles");
-      addPluginStatusIfRegistered(plugins, "decision_logs");
-      addPluginStatusIfRegistered(plugins, "status");
-      addPluginStatusIfRegistered(plugins, "discovery");
+      for (String pluginName : manager.getPluginNames()) {
+        addPluginStatusIfRegistered(plugins, pluginName);
+      }
       report.set("plugins", plugins);
 
       return report;
--- a/opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/StatusPlugin.java
+++ b/opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/StatusPlugin.java
@@ -183,9 +183,14 @@ public final class StatusPlugin implements Plugin {
       }
       report.set("bundles", bundles);
 
-      // Add plugin statuses (only for registered plugins)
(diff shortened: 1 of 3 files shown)


opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/StatusPlugin.java
```diff
@@ -185,10 +185,9 @@ private ObjectNode buildStatusReport() {
 
       // Add plugin statuses (only for registered plugins)
       ObjectNode plugins = MAPPER.createObjectNode();
-      addPluginStatusIfRegistered(plugins, "bundles");
-      addPluginStatusIfRegistered(plugins, "decision_logs");
-      addPluginStatusIfRegistered(plugins, "status");
-      addPluginStatusIfRegistered(plugins, "discovery");
+      for (String pluginName : manager.getPluginNames()) {
```
sspaink, line 190: This loop now includes **every** registered plugin, which pulls in `services` — `ServicePlugin` is registered under the key `"services"` (`Opa.java:781` no-discovery, `Opa.java:750` discovery) and sets `updatePluginStatus("services", OK)` in `start()` (`ServicePlugin.java:239`). So every status report now gains a `"services":"OK"` entry in the `plugins` object.

That's a behavior change from the previous hardcoded list, which deliberately reported only `bundles`, `decision_logs`, `status`, and `discovery` and left `services` out. It also diverges from Go OPA, where REST service clients aren't status-reporting plugins and don't appear in the status report's `plugins` block.

Could you filter `services` out so the report matches the prior behavior and OPA? For example:

```java
for (String pluginName : manager.getPluginNames()) {
  if ("services".equals(pluginName)) {
    continue;
  }
  addPluginStatusIfRegistered(plugins, pluginName);
}
```

The custom-plugin auto-discovery this PR adds is great — we just don't want `services` leaking into the report alongside it.

Develop-KIM: Done — filtered out `services` and added a regression test asserting it stays out of the report while custom plugins remain. Thanks for the catch!

sspaink (approved):
Thank you!

PR #151, opened 2026-07-11
Implement providers.aws.sign_req builtin

### LINKED ISSUE #146: Implement providers.aws.sign_req builtin
## Missing builtin

- `providers.aws.sign_req`

## Context

This OPA builtin is not implemented in the SDK. It is exercised by the Rego compliance suite and currently raises `FunctionNotFoundError`.

Signs an HTTP request with AWS SigV4.

## Implementation notes

Register with an `@OpaBuiltin`-annotated static method (see `io.github.open_policy_agent.opa.ast.builtin.impls`). Core builtins are in `BuiltinRegistry.BUILTIN_CLASSES`; domain-specific ones live in an `opa-builtins-*` module via the `BuiltinProvider` SPI.

ume3445:
## Summary

Implements providers.aws.sign_req (AWS SigV4 header-based request signing) as described in #146.

## Changes

- Added a new opa-builtins-providers-aws module following the existing domain-specific builtin pattern (BuiltinProvider SPI, similar to opa-builtins-crypto and opa-builtins-time)
- Implements the full SigV4 flow: canonical request construction, string-to-sign, and the HMAC-SHA256 key derivation chain (kSecret -> kDate -> kRegion -> kService -> kSigning)
- Matched against OPA's reference Go implementation (topdown/providers.go, internal/providers/aws/signing_v4.go) rather than reverse-engineered from the fixtures alone
- Validates required request keys (method, url) and AWS config keys (aws_access_key, aws_region, aws_secret_access_key, aws_service), with exact error message wording matched to the compliance fixtures, including alphabetically-sorted key lists in error messages
- Handles raw_body taking precedence over body, disable_payload_signing using the literal UNSIGNED-PAYLOAD and ignoring the body, existing x-amz-content-sha256 headers being overwritten, and signed-headers list construction (lowercased, sorted, including host/x-amz-content-sha256/x-amz-date/x-amz-security-token as applicable)

## Testing

- Verified against all 18 compliance fixture cases (12 success, 6 error) from providers-aws/aws-sign_req.json and aws-sign_req-errors.json
- Cross-checked computed signatures against expected fixture values for all 12 success cases individually
- Full opa-evaluator suite shows no new regressions (confirmed by stashing the change and reproducing the identical pre-existing failure count)

Closes #146.


opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
```diff
@@ -0,0 +1,540 @@
...
+        case '\n':
+          sb.append("\\n");
+          break;
+        case '\r':
+          sb.append("\\r");
+          break;
+        case '\t':
+          sb.append("\\t");
+          break;
+        default:
```
sspaink, line 613: **`body` JSON serialization omits Go's HTML escaping, producing a wrong content hash.** OPA serializes `body` with `json.Marshal`, which HTML-escapes `&`→`\u0026`, `<`→`\u003c`, `>`→`\u003e` (and U+2028/U+2029). This `writeJsonString` emits them literally. Confirmed by running Go: `{"q":"a&b<c>d"}` marshals to `{"q":"a\u0026b\u003cc\u003ed"}` (Go) vs `{"q":"a&b<c>d"}` (here). Any `body` string containing `&`/`<`/`>` (very common) → different bytes → different `x-amz-content-sha256` → wrong signature. Reachable via the `body` path; the fixtures use integer-only bodies so they don't catch it. Add cases for `&`/`<`/`>` (and U+2028/U+2029) to `writeJsonString`.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
```diff
@@ -0,0 +1,540 @@
...
+  private static URI parseUrl(String value) {
+    try {
+      return new URI(value);
+    } catch (URISyntaxException e) {
+      throw new TypeError("providers.aws.sign_req: could not parse url: " + value);
+    }
+  }
+
+  private static String hostFromUrl(URI url) {
+    String host = url.getHost();
```
sspaink, line 414: **`URI.getHost()` drops the port, but OPA signs `url.Host` including the port.** OPA sets `awsHeaders["host"] = theURL.Host`, which is `example.com:8443` for a ported URL; `getHost()` returns just `example.com`. Confirmed via Go: `url.Parse("https://example.com:8443/p").Host == "example.com:8443"`. Since `host` is a signed header, any non-default-port endpoint (LocalStack, MinIO, VPC interface endpoints, custom/test endpoints) yields a different canonical request → wrong signature; IPv6 is affected too. Fixtures use portless URLs. Reconstruct host as `getHost()` + `:` + `getPort()` when a port is present (matching `url.Host`). Related: `parseUrl` uses `new URI`, which is stricter than Go's `url.Parse` and throws on inputs like spaces that OPA would sign.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
```diff
@@ -0,0 +1,540 @@
...
+    } else if (value instanceof RegoInt32) {
+      sb.append(((RegoInt32) value).getValue());
+    } else if (value instanceof RegoBigInt) {
+      sb.append(((RegoBigInt) value).getValue().toString());
+    } else if (value instanceof RegoDecimal) {
+      double d = ((RegoDecimal) value).getValue();
+      if (d == Math.floor(d) && !Double.isInfinite(d)) {
+        sb.append(Long.toString((long) d));
+      } else {
+        sb.append(Double.toString(d));
```
sspaink: **Float `body` values serialize via `Double.toString`, diverging from Go's number formatting.** OPA marshals numbers from a `json.Number`, preserving the compact JSON form (e.g. `1e21`), whereas `Double.toString(1e21)` yields `1.0E21` (always forces a decimal point, uppercase `E`). Confirmed Go emits `1e21`. Non-integer or large-magnitude numbers anywhere in `body` → wrong bytes → wrong signature. Integer values are unaffected, so fixtures pass.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
```diff
@@ -0,0 +1,540 @@
...
+      if (!(headersTerm instanceof RegoObject)) {
+        throw new TypeError("operand 0 must be object but got " + headersTerm.getTypeName());
+      }
+      headersObj = (RegoObject) headersTerm;
+    }
+
+    // disable_payload_signing (optional boolean).
+    boolean disablePayloadSigning = false;
+    RegoValue dps = awsConfigObj.getProperty("disable_payload_signing");
+    if (dps != null) {
```
sspaink, line 185: **(minor) Negative `time_ns` is rejected where OPA accepts it.** The `ts < 0` guard throws `TypeError`, but OPA does `time.Unix(0, ts)`, which happily produces a pre-1970 signing time. Divergence only for negative timestamps — unlikely in practice, flagging for completeness.

sspaink (commented):
Verified against OPA's `topdown/providers.go` + `internal/providers/aws/signing_v4.go`, with the divergences below confirmed by running Go directly. The SigV4 chain, string-to-sign, credential scope, header precedence (including the subtle "existing `x-amz-content-sha256` wins in the canonical request but computed hash wins in output"), `raw_body` precedence, and error-message wording all match, and the fixture-driven test is solid. Because signing is byte-exact, the findings all manifest as wrong signatures on inputs the 18 fixtures don't cover — so the suite passes while real requests would fail AWS verification. Findings 1 and 3 share a root cause: `canonicalJson` reimplements Go's `encoding/json` and misses two of its behaviors.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
ume3445: Fixed, added the five missing escape cases (&, <, >, U+2028, U+2029) matching Go's encoding/json output, verified against known Go json.Marshal output for strings and nested objects. New tests in ProvidersAwsCanonicalJsonTest.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
ume3445: Fixed, hostFromUrl now appends :port when present, matching Go's url.Host behavior. Verified with both ported and portless URLs in ProvidersAwsHostFromUrlTest.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
ume3445: Fixed, implemented Go-compatible float formatting (lowercase e, no forced decimal point, correct exponent cleanup). Verified against known Go json.Marshal output for large magnitudes, small decimals, and negative zero in ProvidersAwsCanonicalJsonTest.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
ume3445: Looked into matching time.Unix(0, ts) exactly, but the existing failure-simple-bad timestamp fixture explicitly expects -1e9 to throw an error, so this change breaks that fixture. Since you flagged it as minor, I held off changing shared test data unilaterally, happy to update both the code and the fixture if you'd rather match Go exactly here too.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
```diff
@@ -0,0 +1,608 @@
...
+    for (String key : REQUIRED_REQUEST_KEYS) {
+      if (reqObj.getProperty(key) == null) {
+        missing.add(key);
+      }
+    }
+    if (!missing.isEmpty()) {
+      throw new TypeError(
+          "operand 1 missing required request parameters(s): " + formatKeySet(missing));
+    }
+  }
```
sspaink, line 342: `validateHTTPRequestOperand` also rejects keys outside `http.send`'s `allowedKeyNames` whitelist, and that check runs *before* the missing-keys check ([http.go:311](https://github.com/open-policy-agent/opa/blob/bfd7632b63d6b883102a05bd26568a6ced1b12cf/v1/topdown/http.go#L311)) — so `{"url": ..., "method": ..., "bogus_key": "x"}` errors in OPA but signs successfully here. The whitelist is the 28-entry [`allowedKeyNames`](https://github.com/open-policy-agent/opa/blob/bfd7632b63d6b883102a05bd26568a6ced1b12cf/v1/topdown/http.go#L46-L75) list, and note the invalid-keys error needs to precede the missing-keys one to match OPA's ordering.

sspaink (commented):
Thanks for the quick turnaround on these!

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
sspaink: The fixture isn't rejecting `-1e9` for being negative — OPA's `timestamp.Int64()` is `strconv.ParseInt(literal, 10, 64)`, which rejects any literal that isn't a plain base-10 integer. Confirmed against the binary: `-1e9` errors but `-1000000000` **signs fine** (`19691231T235959Z`), while `1e9` and `1.0` both **error**. So `ts < 0` satisfies the fixture by coincidence and is wrong in both directions — it rejects valid negative integers and accepts `1e9`/`1.0`, which OPA rejects.

The correct gate is the number's literal form rather than its value, which needs `RegoDecimal`/`RegoBigInt` to retain the source text (same root cause as the float-body thread). Fine as a follow-up rather than a blocker — but worth reopening, since the fixture is currently passing for the wrong reason.

opa-builtins/opa-builtins-providers-aws/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ProvidersAwsBuiltins.java
```diff
@@ -0,0 +1,608 @@
...
+  // ------------------------------------------------------------------
+
+  private static String canonicalJson(RegoValue value) {
+    StringBuilder sb = new StringBuilder();
+    writeJson(value, sb);
+    return sb.toString();
+  }
+
+  private static void writeJson(RegoValue value, StringBuilder sb) {
+    if (value == null || value instanceof RegoNull) {
```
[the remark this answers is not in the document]
sspaink: Dismissing this one — the `instanceof` chain over the `RegoValue` subtypes matches how the other builtin impls dispatch, and `switch` pattern matching isn't available non-preview on Java 17. A visitor here would add indirection without making the code clearer.

sspaink (approved):

PR #153, opened 2026-07-13
Implement io.jwt.verify_eddsa builtin (closes #145)

### LINKED ISSUE #145: Implement io.jwt.verify_eddsa builtin
## Missing builtin

- `io.jwt.verify_eddsa`

## Context

This OPA builtin is not implemented in the SDK. It is exercised by the Rego compliance suite and currently raises `FunctionNotFoundError`.

Sibling of the existing `io.jwt.verify_*` builtins; belongs in the `opa-builtins-token` module.

## Implementation notes

Register with an `@OpaBuiltin`-annotated static method (see `io.github.open_policy_agent.opa.ast.builtin.impls`). Core builtins are in `BuiltinRegistry.BUILTIN_CLASSES`; domain-specific ones live in an `opa-builtins-*` module via the `BuiltinProvider` SPI.

ume3445:
## Summary

Implements io.jwt.verify_eddsa as described in #145.

## Changes

- Added a public verifyEdDSA method to TokenBuiltins.java, following the exact pattern of the existing verify_es256/verify_es384/verify_es512 methods
- Delegates to the existing private _verifyEdDSA method, which already correctly implements EdDSA verification matching OPA's Go reference implementation (topdown/tokens.go): raw signing input with no pre-hashing, verified via Ed25519 through BouncyCastle
- Registered the new builtin in the builtins() map alongside the other io.jwt.verify_* entries
- The verification logic itself was already correct and in use internally by io.jwt.decode_verify; this PR only exposes it as a standalone public builtin

## Testing

- Verified against the jwtverifyeddsa/success-cert compliance fixture, which now correctly evaluates and passes (previously the missing function meant the test was silently skipped rather than actually asserted, since the test harness ignores FunctionNotFoundError cases)
- Confirmed the existing jwtdecodeverify EdDSA case still passes, since it exercises the same shared _verifyEdDSA path
- Full opa-evaluator suite shows no new regressions (stash-and-compare against the pre-existing failure set, identical 62 unrelated failures with and without the change)

Closes #145.


opa-builtins/opa-builtins-token/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/TokenBuiltins.java
```diff
@@ -849,6 +850,31 @@ public RegoBoolean verifyES512(EvaluationContext ctx, RegoValue[] args) {
...
+      args = {
+        @OpaType(
+            type = "string",
+            name = "jwt",
+            description = "JWT token whose signature is to be verified"),
+        @OpaType(
+            type = "string",
+            name = "certificate",
+            description =
+                "PEM encoded certificate, PEM encoded public key, or the JWK key (set) used to verify the signature")
```
sspaink, line 912: `_verifyEdDSA` reaches `getPublicKey` → `extractPublicKey`, which only handles `RSAKey` and `ECKey`. An Ed25519 JWK parses to an `OctetKeyPair` (`kty: OKP`), so it falls into the `else` branch and throws `"Unsupported JWK key type"` — meaning this returns `false` (or errors in strict mode) for a valid Ed25519 JWK, even though the arg doc here advertises "the JWK key (set)".

This diverges from OPA: its `getKeysFromCertOrJWK` (`v1/topdown/tokens.go`) parses JWKs via jwx's `jwk.Export`, which supports OKP keys, so OPA accepts an Ed25519 JWK for `io.jwt.verify_eddsa`. For parity, consider adding an OKP branch to `extractPublicKey` (convert `OctetKeyPair` → Ed25519 `PublicKey`) rather than only supporting the PEM path. The added test only covers the PEM cert fixture, so the JWK path is untested for EdDSA.

sspaink (approved):

PR #156, opened 2026-07-14
Implement uri.parse and uri.is_valid builtins

### LINKED ISSUE #135: Implement URI builtins (uri.*)
## Missing builtins

- `uri.parse`
- `uri.is_valid`

## Context

## Implementation notes

shanksmp:
Closes #135

## Summary

- implement `uri.parse`
- implement `uri.is_valid`
- register both URI builtins with the evaluator
- preserve decoded and raw URI path components
- support IPv6 hostnames and optional URI components

## Testing

- URI compliance tests pass
- evaluator compilation passes

The complete evaluator build also encountered unrelated Windows-specific failures in existing time-format and filesystem-path tests.

### DIFF
--- a/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
+++ b/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
@@ -11,6 +11,9 @@ import io.github.open_policy_agent.opa.rego.EvaluationContext;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Locale;
 import java.util.Map;
 import java.util.function.BiFunction;
 
@@ -41,7 +44,7 @@ public class UriBuiltins {
     }
 
     try {
-      new URI(input.getValue());
+      parseGoCompatibleUri(input.getValue());
       return RegoBoolean.TRUE;
     } catch (URISyntaxException e) {
       return RegoBoolean.FALSE;
@@ -58,46 +61,32 @@ public class UriBuiltins {
     String input = getArg(args, 0, RegoString.class).getValue();
 
     try {
-      URI uri = new URI(input);
+      ParsedUri uri = parseGoCompatibleUri(input);
       RegoObject result = new RegoObject();
 
-      if (uri.getScheme() != null && !uri.getScheme().isEmpty()) {
-        result.setProperty("scheme", new RegoString(uri.getScheme()));
+      if (uri.scheme() != null) {
+        result.setProperty("scheme", new RegoString(uri.scheme()));
       }
 
-      String hostname = uri.getHost();
-      if (hostname != null && !hostname.isEmpty()) {
-        if (hostname.startsWith("[") && hostname.endsWith("]")) {
-          hostname = hostname.substring(1, hostname.length() - 1);
-        }
-
-        result.setProperty("hostname", new RegoString(hostname));
+      if (uri.hostname() != null) {
+        result.setProperty("hostname", new RegoString(uri.hostname()));
       }
 
-      if (uri.getPort() != -1) {
-        result.setProperty("port", new RegoString(Integer.toString(uri.getPort())));
+      if (uri.port() != null) {
+        result.setProperty("port", new RegoString(uri.port()));
       }
 
-      String path = uri.getPath();
-      if (path != null && !path.isEmpty()) {
-        result.setProperty("path", new RegoString(path));
-
-        String rawPath = uri.getRawPath();
-        if (rawPath == null || rawPath.isEmpty()) {
-          rawPath = path;
-        }
-
-        result.setProperty("raw_path", new RegoString(rawPath));
+      if (uri.path() != null) {
+        result.setProperty("path", new RegoString(uri.path()));
+        result.setProperty("raw_path", new RegoString(uri.rawPath()));
       }
 
-      String rawQuery = uri.getRawQuery();
-      if (rawQuery != null && !rawQuery.isEmpty()) {
-        result.setProperty("raw_query", new RegoString(rawQuery));
+      if (uri.rawQuery() != null) {
+        result.setProperty("raw_query", new RegoString(uri.rawQuery()));
       }
 
-      String fragment = uri.getFragment();
-      if (fragment != null && !fragment.isEmpty()) {
-        result.setProperty("fragment", new RegoString(fragment));
+      if (uri.fragment() != null) {
+        result.setProperty("fragment", new RegoString(uri.fragment()));
       }
 
       return result;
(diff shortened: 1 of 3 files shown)


opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
```diff
@@ -0,0 +1,108 @@
...
+
+    try {
+      URI uri = new URI(input);
+      RegoObject result = new RegoObject();
+
+      if (uri.getScheme() != null && !uri.getScheme().isEmpty()) {
+        result.setProperty("scheme", new RegoString(uri.getScheme()));
+      }
+
+      String hostname = uri.getHost();
```
sspaink: **Hostname and port are silently dropped for registry-based authorities.** `java.net.URI.getHost()` returns `null` whenever the authority isn't a strict RFC-2396 server-based host (e.g. a hostname containing an underscore), and `getPort()` then returns `-1`. For `uri.parse("http://my_host:8080/path")` this returns `{scheme:"http", path:"/path", raw_path:"/path"}` — hostname and port missing — whereas OPA returns `{scheme:"http", hostname:"my_host", port:"8080", path:"/path", raw_path:"/path"}`. Underscores are common in internal/Docker/service hostnames, and `http://my_host.example.com/` triggers it too. Consider falling back to parsing `uri.getAuthority()` when `getHost()` is null.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
```diff
@@ -0,0 +1,108 @@
...
+      args = {@OpaType(name = "x", description = "URI string to parse")},
+      result = @OpaType(name = "result", description = "Parsed URI components"))
+  public RegoObject parse(EvaluationContext ctx, RegoValue[] args) {
+    String input = getArg(args, 0, RegoString.class).getValue();
+
+    try {
+      URI uri = new URI(input);
+      RegoObject result = new RegoObject();
+
+      if (uri.getScheme() != null && !uri.getScheme().isEmpty()) {
```
sspaink: **`scheme` is not lowercased, unlike OPA.** Go's `url.Parse` lowercases the scheme, so `uri.parse("HTTP://EXAMPLE.COM")` returns `scheme:"http"` in OPA but `scheme:"HTTP"` here (`URI.getScheme()` preserves case). Lower impact, but a real output difference for any mixed-case scheme.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
```diff
@@ -0,0 +1,108 @@
...
+      name = "uri.parse",
+      description = "Parses an input URI and returns its components.",
+      categories = {"encoding"},
+      args = {@OpaType(name = "x", description = "URI string to parse")},
+      result = @OpaType(name = "result", description = "Parsed URI components"))
+  public RegoObject parse(EvaluationContext ctx, RegoValue[] args) {
+    String input = getArg(args, 0, RegoString.class).getValue();
+
+    try {
+      URI uri = new URI(input);
```
sspaink: **`uri.parse` throws on inputs OPA parses successfully.** `new URI(...)` rejects chars that are legal to `net/url.Parse`, e.g. `uri.parse("https://example.com/path with spaces")` or `.../p|pe` throw `URISyntaxException` → `BuiltinError`, whereas OPA returns a normal object with the raw path (`path:"/path with spaces"`). This turns a successful evaluation into an error.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
```diff
@@ -0,0 +1,108 @@
...
+    if (!(args[0] instanceof RegoString input)) {
+      return RegoBoolean.FALSE;
+    }
+
+    if (input.getValue().isEmpty()) {
+      return RegoBoolean.FALSE;
+    }
+
+    try {
+      new URI(input.getValue());
```
sspaink: **`uri.is_valid` returns `false` for inputs OPA considers valid.** Same `java.net.URI` strictness as in `parse`: `uri.is_valid("https://example.com/path with spaces")` and `uri.is_valid("http://example.com/p|pe")` return `false` here, but Go's `url.Parse` accepts both so OPA returns `true`.

sspaink (commented):
Output shape (port-as-string, `raw_path` = `cmp.Or(rawPath, path)`, omitted empty components, empty-string→`false` and non-string handling) all correctly match OPA's reference `v1/topdown/uri.go`, and the happy-path compliance cases pass. However, the parser choice — `java.net.URI` (RFC 2396, strict, server-based authority) — diverges from Go's `net/url.Parse` (lenient) on common inputs the fixture doesn't exercise, so the builtins silently produce results that differ from OPA. Details inline. All share one root cause, so the fix is best made at the parser level (e.g. fall back to `getAuthority()` parsing when `getHost()` is null, lowercase the scheme, and use a parser that mirrors `net/url` acceptance). Worth extending the compliance fixture with an underscore-host case and a special-char case to catch these.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
shanksmp: Fixed in the latest commit. URI authority parsing no longer depends on java.net.URI.getHost(). The authority is parsed directly, so registry-based hostnames such as my_host and their ports are preserved. I also added a regression test for http://my_host:8080/path, verifying that hostname is my_host and port is "8080".

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
shanksmp: Fixed in the latest commit. The parsed scheme is now normalized to lowercase using Locale.ROOT, matching OPA’s behavior. I also added a regression test for uri.parse("HTTP://EXAMPLE.COM") that verifies the returned scheme is "http".

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
shanksmp: Fixed in the latest commit. uri.parse now uses the shared OPA-compatible parsing logic instead of passing the input directly to new URI(...). It accepts literal spaces and pipes in paths and returns the expected path and raw_path. Regression tests were added for both cases.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/UriBuiltins.java
shanksmp: Fixed in the latest commit. uri.is_valid now uses the same shared OPA-compatible parsing logic as uri.parse, so inputs containing literal spaces or pipes in their paths are accepted consistently. Regression tests were added for both cases.

shanksmp:
Post this general comment after replying to the inline comments:
I’ve addressed the reported compatibility differences in the latest commit.
Instead of adding only a getAuthority() fallback, I introduced shared parsing logic used by both uri.parse and uri.is_valid, because the underlying java.net.URI strictness affected multiple behaviors. The update covers:

- lowercasing mixed-case schemes
- preserving registry-based hostnames and ports
- accepting literal spaces and pipes in paths
- keeping uri.parse and uri.is_valid consistent
- rejecting control characters and malformed percent escapes

I added focused UriBuiltinsTest regression tests rather than modifying the generated compliance fixture directly. The new regression tests and the existing URI compliance tests pass.
This change targets the reported differences with OPA’s URI behavior; I have not assumed that it reproduces every edge case of Go’s net/url parser. Please let me know if you would prefer these cases to also be added to the compliance fixture.

sspaink (approved):
Great work, thanks!

Issue #137, opened 2026-07-10
Implement union builtins (union, object.union)


sspaink:
## Missing builtins

- `union`
- `object.union`

## Context

These OPA builtins are not implemented in the SDK. Most are exercised by the Rego compliance suite (`opa-evaluator/src/test/resources/compliance/.../TestData`) and currently raise `FunctionNotFoundError`; implementing them makes those cases pass.

## Implementation notes

Register each builtin with an `@OpaBuiltin`-annotated static method (see `io.github.open_policy_agent.opa.ast.builtin.impls`). Core builtins are listed in `BuiltinRegistry.BUILTIN_CLASSES`; domain-specific ones live in an `opa-builtins-*` module exposed via the `BuiltinProvider` SPI (e.g. `RegexBuiltins`, `CidrBuiltins`, `SemverBuiltins`).

Reference: https://www.openpolicyagent.org/docs/policy-reference/#built-in-functions

ML-dev-crypto:
Hi I'm interested in working on this issue and will submit a PR once I've completed the implementation. Could you please assign it to me? Thank you.

PR #157, opened 2026-07-14
Fix missing/incorrect @OpaBuiltin annotations for union and object.union

ML-dev-crypto:
## Summary
Both `union` and `object.union` are already fully implemented and functionally
correct  registered in their respective `builtins()` maps and dispatched
properly at runtime, so calling either from a Rego policy already works today.
The actual gap is in their `@OpaBuiltin` metadata annotations, which are used
to generate capabilities.json and docs. This PR fixes that metadata rather
than reimplementing the builtins, so the diff is smaller than the issue's
"missing builtins" framing might suggest.

## Issues
#137

## Changes
- Added the missing `@OpaBuiltin` annotation to `union` (SetBuiltins.java).
  It previously had none, so it was falling back to a minimal
  auto-generated capabilities entry with no description or arg types.
- Replaced `object.union`'s `@OpaBuiltin` annotation (ObjectBuiltins.java).
  It was copy-pasted from `object.keys` wrong name, wrong single-arg
  signature, wrong description. This also meant two methods in the same
  file both declared `name = "object.keys"`, an ambiguous collision since
  annotation-scanning order over declared methods isn't guaranteed by the
  JVM. Fixing the name resolves the collision as a side effect.
- Swapped a raw `(RegoSet) args[0]` cast for the existing `getArg` helper
  in `union`, so a bad input type throws a `TypeError` instead of a raw
  `ClassCastException` consistent with `intersection`/`and`/`or` in the
  same file.

## Testing
- `./gradlew :opa-evaluator:test` — full suite passes, including all
  compliance cases under `TestData/v1/union/` and `TestData/v1/objectunion/`.
- `./gradlew checkstyleMain checkstyleTest pmdMain pmdTest` — clean, no
  violations in either changed file.
- 11 pre-existing, unrelated failures were present both with and without
  this change (verified via `git stash`): 9 in `FileSystemBundleLoaderTest`
  (Windows-specific `InvalidPathException`, this repo's CI only runs
  `ubuntu-latest`) and 2 in `GoTimeLayoutConverterTest` (unrelated to
  bundle/builtin code).

### DIFF
--- a/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ObjectBuiltins.java
+++ b/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ObjectBuiltins.java
@@ -65,20 +65,34 @@ public class ObjectBuiltins {
   }
 
   @OpaBuiltin(
-      name = "object.keys",
-      description = "Returns a set of keys in the supplied object.",
+      name = "object.union",
+      description =
+          "Creates a new object of the asymmetric union of two objects. For example: "
+              + "`object.union({\"a\": 1}, {\"b\": 2})` results in `{\"a\": 1, \"b\": 2}`. "
+              + "If both objects have the same key and both values are objects, the values are "
+              + "merged recursively. Otherwise, the value in the right-hand object `b` wins.",
+      categories = {"objects"},
       args = {
         @OpaType(
             type = "object",
-            name = "object",
-            description = "object to get keys from",
+            name = "a",
+            description = "left-hand object",
+            dynamic = @OpaDynamic(keyType = "any", valueType = "any")),
+        @OpaType(
+            type = "object",
+            name = "b",
+            description = "right-hand object",
             dynamic = @OpaDynamic(keyType = "any", valueType = "any"))
       },
       result =
           @OpaType(
-              type = "set",
-              description = "set of keys in `object`",
-              dynamic = @OpaDynamic(type = "any")))
+              type = "object",
+              name = "output",
+              dynamic = @OpaDynamic(keyType = "any", valueType = "any"),
+              description =
+                  "a new object which is the result of an asymmetric recursive union of two "
+                      + "objects where conflicts are resolved by choosing the key from the "
+                      + "right-hand object `b`"))
   public RegoObject union(EvaluationContext ctx, RegoValue[] args) {
     if (!(args[0] instanceof RegoObject)) {
       throw new TypeError("object.union: operand 1 must be object but got " + args[0].getTypeName());
--- a/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/SetBuiltins.java
+++ b/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/SetBuiltins.java
@@ -26,8 +26,25 @@ public class SetBuiltins {
         "union", instance::union);
   }
 
+  @OpaBuiltin(
+      name = "union",
+      description = "Returns the union of the given input sets.",
+      categories = {"sets"},
+      args = {
+        @OpaType(
+            type = "set",
+            name = "xs",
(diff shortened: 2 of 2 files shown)


opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/ObjectBuiltins.java
```diff
@@ -65,20 +65,33 @@ public RegoObject unionN(EvaluationContext ctx, RegoValue[] args) {
...
+            name = "b",
+            description = "right-hand object",
             dynamic = @OpaDynamic(keyType = "any", valueType = "any"))
       },
       result =
           @OpaType(
-              type = "set",
-              description = "set of keys in `object`",
-              dynamic = @OpaDynamic(type = "any")))
+              type = "any",
```
sspaink: **`object.union` result is declared `type = "any"`, but OPA declares it as an `object` with dynamic key/value — so the generated capability still diverges from OPA.** Running `CapabilitiesGenerator` on this branch emits:
```json
"result": { "type": "any", "name": "output", ... }
```
but OPA's `capabilities.json` for `object.union` result is:
```json
{ "type": "object", "dynamic": { "key": { "type": "any" }, "value": { "type": "any" } } }
```
The method returns `RegoObject`, and the two args are already correctly annotated as `object` + dynamic — only the result was left as `any`. Since matching `capabilities.json` is the whole point of this PR, this leaves the headline fix incomplete. Should be `type = "object", dynamic = @OpaDynamic(keyType = "any", valueType = "any")`.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/SetBuiltins.java
```diff
@@ -26,8 +26,25 @@ public static Map<String, BiFunction<EvaluationContext, RegoValue[], RegoValue>>
...
+  @OpaBuiltin(
+      name = "union",
+      description = "Returns the union of the given input sets.",
+      categories = {"sets"},
+      args = {
+        @OpaType(
+            type = "set",
+            name = "xs",
+            description = "set of sets to merge",
+            dynamic = @OpaDynamic(type = "set", keyType = "any", valueType = "any"))
```
sspaink: **`@OpaDynamic(type = "set", keyType = "any", valueType = "any")` mixes mutually-exclusive fields; `keyType`/`valueType` are silently dropped.** `OpaDynamic`'s contract states `type` is "mutually exclusive with key/value," and `CapabilitiesGenerator.convertOpaTypeToArgument` checks `dynamic.type()` first and returns — so the emitted arg on this branch is just:
```json
"dynamic": { "type": "set" }
```
The `keyType`/`valueType` have no effect and are misleading (key/value shape is for objects, not sets). Suggest dropping them to `@OpaDynamic(type = "set")`, matching how the sibling set builtins (`and`/`or`) annotate their set args.

(Separately and out of scope here: this repo expresses set element types via `dynamic` while OPA uses `of`, so the `set` shape won't byte-match OPA regardless — but that's a pre-existing convention across `and`/`or`/`intersection`, not something this PR introduces.)

sspaink (commented):
Metadata-only PR — verified against OPA's `capabilities.json` / `v1/ast/builtins.go` and by running this repo's `CapabilitiesGenerator` on this branch and diffing the emitted JSON for both builtins. The collision fix is correct (`object.keys` keeps its own annotation on the `keys` method) and the `getArg` swap is fine. Two annotation-shape issues remain; both confirmed against the actually-generated capabilities output.

ML-dev-crypto:
Thanks for the detailed review! Good catch on both points. I've updated the object.union result annotation to use type = "object" with the appropriate dynamic key/value types, and simplified the union argument annotation to @OpaDynamic(type = "set") to match the existing convention.

sspaink (approved):
Thanks!

PR #169, opened 2026-07-20
Implemented the poll chain to await the download future that serialize polling

### LINKED ISSUE #113: BundleDownloader: async download allows overlapping polls → concurrent activateBundle/Store mutation
## Background

Follow-up from #71, which converted `BundleDownloader.handleHttpDownload` from a synchronous `httpClient.send(...)` to `httpClient.sendAsync(...).whenComplete(...)`. This introduced a concurrency regression that was considered out of scope for that security-hardening PR.

## Problem

Previously `downloadBundle()` blocked inside `httpClient.send(...)`, so the poll chain (`scheduleNextPoll` schedules the next tick in its `finally`, *after* `downloadBundle()` returns) guaranteed a download fully completed before the next one began — downloads were serialized.

Now `downloadBundle()` returns immediately after `sendAsync`, and the completion callback runs on the HTTP client's executor thread, not the scheduler thread. If a download takes longer than the poll delay (large bundle / slow server), the next poll fires `downloadBundle()` while the previous request is still in flight. The two `whenComplete` callbacks then run `activateBundle()` concurrently, mutating the shared `Store` in parallel.

This affects both:
- `BundlePlugin.Bundle` — concurrent `TarballBundleLoader.load()` into the shared `Store`
- `DiscoveryPlugin.DiscoveryBundle` (inherits the base class) — concurrent `manager.updateConfig(...)`

## How Go OPA handles this

The Go implementation runs a single-goroutine loop that is strictly serial (`v1/download/download.go`):

```go
func (d *Downloader) loop(ctx context.Context) {
    for {
        err := d.oneShot(ctx)   // blocks through download AND the activation callback
        // ... compute jittered delay ...
        select {
        case <-timer.C:         // wait the delay only AFTER oneShot fully returns
        case <-ctx.Done():
            return
        }
    }
}
```

The jittered delay is applied only *after* `oneShot` (download + callback) completes, so downloads never overlap. The Java `scheduleNextPoll` comment explicitly says it is "matching Go-OPA's jittered polling"; the old synchronous code preserved that serial guarantee, and the `sendAsync` rewrite broke it.

## Suggested fix

Serialize activation with the poll chain again. Options:
- Have the poll chain await the download future before scheduling the next tick, or
- Confine activation to a single thread (e.g. hop back onto the scheduler thread in `whenComplete`).

## Acceptance

- Overlapping downloads can no longer trigger concurrent `activateBundle()` calls.
- A test that simulates a slow download exceeding the poll interval confirms downloads do not overlap.

polachandu:
Re-serialize by chaining each poll off the previous download's completion instead of a fixed timer — the same "single serial loop" guarantee that Go-OPA has. This makes bundle (and discovery) polling strictly serial, so overlapping polls can no longer run activateBundle() — and thus mutate the shared Store — concurrently.

Added a test confirming that downloads remain serialized even when activation exceeds the poll interval, so overlapping downloads cannot trigger concurrent activateBundle() calls.

Closes #113

### DIFF
--- a/opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
+++ b/opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
@@ -189,39 +189,45 @@ public abstract class BundleDownloader {
             ? polling.getMaxDelaySeconds()
             : 120;
 
-    scheduler.schedule(this::downloadBundle, 0, TimeUnit.SECONDS);
-    scheduleNextPoll(scheduler, minDelay, maxDelay);
+    // Run the first download immediately, then chain each subsequent poll only after the previous
+    // download AND its (possibly async) activation have fully completed. Chaining off completion
+    // instead of a fixed timer keeps polling strictly serial, so overlapping downloads can never
+    // run activateBundle() concurrently against a shared Store (issue #113).
+    scheduleDownload(scheduler, minDelay, maxDelay, 0);
 
     return initialActivation;
   }
 
-  // Re-schedules the next download with a uniformly random delay in [minDelay, maxDelay],
-  // matching Go-OPA's jittered polling. ScheduledExecutorService has no built-in jitter, so the
-  // task chains itself. RejectedExecutionException after a shutdown breaks the chain cleanly.
-  private void scheduleNextPoll(ScheduledExecutorService scheduler, int minDelay, int maxDelay) {
-    long delay =
-        minDelay >= maxDelay
-            ? minDelay
-            : ThreadLocalRandom.current().nextLong(minDelay, (long) maxDelay + 1);
+  // Schedules one download after delaySeconds; once that download (including any async activation)
+  // completes, schedules the next poll with a fresh jittered delay in [minDelay, maxDelay]. Because
+  // the next poll is chained off the download's completion — not fired on an independent timer —
+  // at most one download is ever in flight, mirroring Go-OPA's strictly serial poll loop.
+  // ScheduledExecutorService has no built-in jitter, so the task chains itself; a
+  // RejectedExecutionException after shutdown ends the chain cleanly.
+  private void scheduleDownload(
+      ScheduledExecutorService scheduler, int minDelay, int maxDelay, long delaySeconds) {
     try {
       scheduler.schedule(
-          () -> {
-            try {
-              downloadBundle();
-            } catch (Exception e) {
-              // downloadBundle() handles its own logging; swallow so the chain keeps polling.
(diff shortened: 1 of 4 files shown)


polachandu:
Hi @sspaink, please review this PR. Please post your comments/suggestions if any, happy to incorporate.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/BundleDownloader.java
```diff
@@ -189,48 +189,54 @@ public CompletableFuture<Void> startPolling(ScheduledExecutorService scheduler)
...
-              // executor's uncaught-exception handler tear down the pool, which is the right
-              // outcome for unrecoverable conditions.
-            } finally {
-              scheduleNextPoll(scheduler, minDelay, maxDelay);
-            }
-          },
-          delay,
+          () ->
+              downloadBundle()
+                  .whenComplete(
```
sspaink, line 212: The request has no `.timeout(...)` and the clients only set `connectTimeout`. Java's `HttpClient` has no default response timeout, so a server that accepts and never replies leaves `sendAsync` pending forever: `whenComplete` never fires, `scheduleDownload` is never called again, and polling stops permanently with nothing logged.

One line next to `requestBuilder.build()` fixes it:

```java
requestBuilder.timeout(Duration.ofSeconds(/* e.g. the service's responseHeaderTimeoutSeconds */));
```

The timeout fails the future, `handleHttpResponse` logs it, and the chain keeps polling. (Prefer this over `.orTimeout(...)`, which doesn't cancel the request — the stalled exchange could still land later and call `activateBundle` alongside a fresh poll.)

polachandu: Looking into it.

sspaink: Ah, my bad — `requestBuilder.timeout()` covers the whole exchange, not just headers, so wiring it to a 10s value would've killed any download slower than that. Real problem, and my suggestion walked you into it.

I'd rather not add `download_timeout_seconds` though. OPA doesn't have a bundle download deadline at all, and `response_header_timeout_seconds` is headers-only — "does not include the time to read the response body." The catch is the buffering body handler, not the timeout: switch to `BodyHandlers.ofInputStream()` and the future completes as soon as headers land, so the deadline only covers headers and a slow body just keeps streaming. No 300s default needed. Bonus, cancel the original future on timeout and my `orTimeout` gripe goes away too.

sspaink (approved):

PR #174, opened 2026-07-21
feat(services): support dynamic entrypoint evaluation via DecisionOptions

babayanv:
### Description
This PR introduces the ability to dynamically override the policy entrypoint on a per-request basis using `DecisionOptions`. 

### Motivation
When working with compiled bundles containing multiple entrypoints (e.g., built using `opa build -t plan -e example/allow -e example/foo -e example/bar -b . -o bundle.tar.gz`), the current `Opa.java` forces all evaluations through the `defaultEntrypoint`.

By adding an `entrypoint` field to `DecisionOptions`, it will now be possible to evaluate different branches of a plan bundle on the fly using a single `Opa` instance.

### Changes
1. **`Opa.DecisionOptions`**: Added a private `entrypoint` field along with its respective getter and setter.
2. **`Opa.buildContext(DecisionOptions options)`**: Updated the evaluation context building block to look up `options.getEntrypoint()`. If provided, it overrides the standard `defaultEntrypoint`.

### DIFF
--- a/opa-services/src/main/java/io/github/open_policy_agent/opa/Opa.java
+++ b/opa-services/src/main/java/io/github/open_policy_agent/opa/Opa.java
@@ -243,17 +243,26 @@ public class Opa {
     boolean useMetrics = showMetrics || (options != null && options.showMetrics);
     Metrics metrics = useMetrics ? new SimpleMetrics() : NoOpMetrics.Instance();
 
+    String entrypoint = resolveEntrypoint(options);
+
     EvaluationContext.Builder builder =
         new EvaluationContext.Builder()
             .withStore(store)
             .withMetrics(metrics)
-            .withEntrypoint(defaultEntrypoint);
+            .withEntrypoint(entrypoint);
 
     if (options != null && options.getProfiler() != null) {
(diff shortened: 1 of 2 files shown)


opa-services/src/main/java/io/github/open_policy_agent/opa/Opa.java
sspaink: Thanks for this! This existing path field was only being used to send the entrypoint to the decision logs. Since your new entrypoint field now represents the same thing, could you remove path (field + getPath/setPath) and have this log line use the resolved entrypoint instead? That way there aren't two fields for one concept, and the log always matches whatever entrypoint actually ran.

babayanv: Hi! Thank you so much for taking the time to review this!

I have removed the path field and its respective getPath() and setPath() methods. I also updated the resolveEntrypoint() method to match the original path evaluation logic for decision logging.

Please let me know if there is anything else I can do!

sspaink (approved):
Thanks!

PR #186, opened 2026-08-06
feat(decision-logs): apply the mask_decision policy to events

### LINKED ISSUE #73: Implement mask_decision policy
`DecisionLogPlugin.DecisionLogs.logDecision` carries a `// TODO: Apply mask decision policy if configured` (around DecisionLogPlugin.java:298). The config field and setter exist, but nothing reads the configured Rego policy or redacts the event. This blocks any production deployment where input or result contains sensitive data.

* Evaluate the configured `mask_decision` policy against the event before buffering.
* Apply the returned redactions per the OPA spec (`{"op":"upsert","path":"/input/foo","value":"<redacted>"}` etc.) — remove, replace, or upsert fields in input / result.
* Pass evaluation errors through the plugin's logger and continue logging the un-masked event (or drop, depending on what OPA Go does — verify before implementing).
* Tests covering remove / upsert / no-op masks and policy evaluation failure.

sspaink:
Fixes #73

Implements `decision_logs.mask_decision`: decision events are redacted before they are buffered, uploaded, or written to the console. `MaskRuleSet` ports `v1/plugins/logs/mask.go`, and a masking failure drops the event rather than logging it unmasked.

The mask policy has to be built as a plan entrypoint (`opa build -t plan -e system/log/mask`), since this SDK evaluates compiled plans. A configured entrypoint with no plan logs a warning.


opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/DecisionLogPlugin.java
```diff
@@ -361,9 +383,12 @@ public void logDecision(
             buildDecisionEvent(
                 decisionId, input, result, path, requestedBy, timestamp, metrics, ndCacheValues);
 
-        // TODO: Apply mask decision policy if configured
         // TODO: Apply drop decision policy if configured
 
+        if (!applyMask(event)) {
+          return; // masking failed: drop the event rather than log it unmasked, as OPA Go does
```
kroekle, line 389: I think this is good for now, but maybe we need to start a conversation on the Go OPA side.  This seems to me to be that should be configurable.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/MaskRuleSet.java
```diff
@@ -0,0 +1,335 @@
...
+          return false;
+        }
+      }
+
+      String target = parts[parts.length - 1];
+      if (node.isObject()) {
+        return ((ObjectNode) node).remove(target) != null;
+      }
+      if (node.isArray()) {
+        // Deliberate divergence: for an array-valued field ("/input/1" with an array input), Go
```
kroekle, line 174: Maybe we should add this into the readme somewhere as well.

opa-services/src/test/java/io/github/open_policy_agent/opa/plugins/DecisionLogPluginMaskTest.java
```diff
@@ -0,0 +1,324 @@
...
+import io.github.open_policy_agent.opa.ir.PolicyReader;
+import io.github.open_policy_agent.opa.logging.Logger;
+import io.github.open_policy_agent.opa.storage.InMem;
+import io.github.open_policy_agent.opa.storage.Store;
+
+/**
+ * Tests for applying the configured {@code mask_decision} policy to decision log events. The mask
+ * policies come from the compiled plans in {@code src/test/resources/mask}.
+ */
+class DecisionLogPluginMaskTest {
```
kroekle, line 28: If I missed it then you can close this, but I think we need a test that does has conflicting rules, say we have a rule to mask and remove an attribute.  I think order is going to be important.

kroekle (commented):
Looks really good.  Maybe just a docs update and a new test and I think it's good to go.

opa-services/src/main/java/io/github/open_policy_agent/opa/plugins/MaskRuleSet.java
sspaink: Added to `opa-services/README.md`

opa-services/src/test/java/io/github/open_policy_agent/opa/plugins/DecisionLogPluginMaskTest.java
sspaink: Tests added 👍

kroekle (approved):
LGTM

PR #197, opened 2026-08-12
feat(evaluator): implement graph builtins

### LINKED ISSUE #132: Implement graph reachability builtins (graph.*)
## Missing builtins

- `graph.reachable`
- `graph.reachable_paths`

## Context

## Implementation notes

dangzitou:
## Summary

- implement graph.reachable and graph.reachable_paths using the traversal semantics from OPA v1/topdown/reachable.go
- register both core builtins with capabilities metadata
- enable the official graph compliance fixtures, including Rego literals with set(), empty objects, and trailing commas
- add focused coverage for array/set edges, cycles, and missing nodes

## Reproduction

On current main, the focused registry tests failed 2/2 on three consecutive runs because both graph builtin names were absent.

## Verification

- ./gradlew :opa-evaluator:test --tests io.github.open_policy_agent.opa.ast.builtin.impls.GraphBuiltinsTest --tests io.github.open_policy_agent.opa.ast.builtin.CapabilitiesGeneratorTest
- ./gradlew :opa-evaluator:pmdMain :opa-evaluator:pmdTest :opa-evaluator:checkstyleMain :opa-evaluator:checkstyleTest
- ./gradlew build -x test
- Full ComplianceTest: every graph fixture passes. The only two remaining failures are time/format and json_match_schema/invalid document; both reproduce unchanged on upstream/main.

Fixes #132

### DIFF
--- a/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/GraphBuiltins.java
+++ b/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/GraphBuiltins.java
@@ -15,11 +15,10 @@ import io.github.open_policy_agent.opa.rego.EvaluationContext;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.function.BiFunction;
 
 public class GraphBuiltins {
@@ -128,7 +127,8 @@ public class GraphBuiltins {
     }
 
     List<RegoValue> path = new ArrayList<>(List.of(initial));
-    Set<RegoValue> reached = new HashSet<>(Set.of(initial));
+    Map<RegoValue, Integer> reached = new HashMap<>();
+    reached.put(initial, 1);
     Deque<PathFrame> pending = new ArrayDeque<>();
     pending.addLast(new PathFrame(null, neighbors));
 
@@ -138,13 +138,18 @@ public class GraphBuiltins {
         pending.removeLast();
         if (frame.node != null) {
           path.remove(path.size() - 1);
-          reached.remove(frame.node);
+          reached.compute(
+              frame.node,
+              (node, visits) -> visits == null || visits == 1 ? null : visits - 1);
         }
         continue;
       }
 
       RegoValue neighbor = frame.neighbors.get(frame.nextNeighborIndex++);
-      if (reached.contains(neighbor)) {
+      // OPA includes a direct self-loop from the initial node, but stops cycles after
+      // entering a child frame. Keep a visit count so unwinding that self-loop does not
+      // remove the initial node from the active path.
+      if (frame.node != null && reached.containsKey(neighbor)) {
         addPath(paths, path);
         continue;
       }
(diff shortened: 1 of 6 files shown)


dangzitou:
Could a maintainer please rerun the failed **PR Check** workflow?

Both attempts failed in **Check what files changed** while `open-policy-agent/setup-opa` was downloading OPA (socket hang up / HTTP 503), before change detection or module tests ran. The other independent checks passed, and the branch is up to date with `main`.

Thanks!

sspaink (commented):
`opa-builtins/README.md:106` still lists `graph.*` as unimplemented, so the table now contradicts this PR. #191 was a dedicated commit to fix this same table drifting, so probably worth folding into this change:

```
| `bits.*`, `units.*`, `http.send` | No |
```

...plus a `Yes` row for `graph.reachable`, `graph.reachable_paths`.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/GraphBuiltins.java
```diff
@@ -0,0 +1,178 @@
...
+
+    if (!ctx.sortSets) {
+      return paths;
+    }
+    List<RegoValue> sortedPaths = new ArrayList<>(paths.getValue());
+    sortedPaths.sort(RegoValue::compareTo);
+    return new RegoSet(false, new LinkedHashSet<>(sortedPaths));
+  }
+
+  private static void buildPaths(
```
sspaink: `buildPaths` recurses once per path node, so deep graphs blow the JVM stack — measured on this branch, a 10k-node chain throws `StackOverflowError` (12k+ gives `OutOfMemoryError`), where OPA Go returns the 20,001-element path fine since goroutine stacks grow.

`Evaluator.invokeBuiltin` only catches `TypeError`/`BuiltinError`, so the `Error` escapes `evaluate()` entirely instead of surfacing as an `OpaException` — and the graph can come from `data`. Worth making this iterative with an explicit work stack, or at minimum bounding depth and throwing `BuiltinError`.

dangzitou: Addressed in f96cf6a: path enumeration now uses an explicit work deque instead of recursive calls, with a 10,000-node chain regression test. The builtin support table now lists `graph.reachable` and `graph.reachable_paths` as supported. Verified with `./gradlew :opa-evaluator:test --tests 'io.github.open_policy_agent.opa.ast.builtin.impls.GraphBuiltinsTest'`.

dangzitou:
Addressed the review feedback in f96cf6a: the builtin support table now lists the graph builtins as supported, and `graph.reachable_paths` uses iterative traversal with a 10,000-node regression test. Local `GraphBuiltinsTest` passes.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ast/builtin/impls/GraphBuiltins.java
```diff
@@ -0,0 +1,192 @@
...
+        pending.removeLast();
+        if (frame.node != null) {
+          path.remove(path.size() - 1);
+          reached.remove(frame.node);
+        }
+        continue;
+      }
+
+      RegoValue neighbor = frame.neighbors.get(frame.nextNeighborIndex++);
+      if (reached.contains(neighbor)) {
```
sspaink: OPA's `builtinReachablePaths` enters `pathBuilder` without a cycle check, so a node that is its own neighbor appears twice in a path — the root frame checks it here, and `graph.reachable_paths({"a": {"a","b"}, "b": set()}, {"a"})` drops the whole `a → a → b` branch: `[["a"], ["a","b"]]` vs `[["a","a"], ["a","a","b"], ["a","b"]]` on `opa eval` 1.19.0.

```suggestion
      if (frame.node != null && reached.contains(neighbor)) {
```

`reached` needs to count occurrences as well, or the pop on line 141 clears the initial node's own entry.

dangzitou: Thank you for catching this edge case. Addressed in caa0660: the root frame now permits the initial self-loop once, while visit counts preserve the root entry during unwind so later back-edges still terminate. Added a regression covering `a -> {a, b}` and verified with `./gradlew :opa-evaluator:test --tests 'io.github.open_policy_agent.opa.ast.builtin.impls.GraphBuiltinsTest'`.

sspaink (approved):
Thank you for the contribution!

PR #201, opened 2026-08-18
tracing: Make coverage range aware

charlieegan3:
Follows functionality added in https://github.com/open-policy-agent/opa/pull/9007 to save this data to IR plans.

Example range based coverage from my local tester (`true, true` yes not make it into the plan, something for the likes of https://github.com/open-policy-agent/opa/pull/9031 etc)

Screenshot

### DIFF
--- a/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/Location.java
+++ b/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/Location.java
@@ -13,10 +13,21 @@ public class Location implements LocationStmt {
 
   private int row; // row in the source file where this statement originated
 
+  private int endCol; // column one past the last rune of this statement's source text
+
+  private int endRow; // row of the last rune of this statement's source text
+
   public Location(int file, int col, int row) {
+    // For callers that do not have end location data.
+    this(file, col, row, col, row);
+  }
+
+  public Location(int file, int col, int row, int endCol, int endRow) {
     this.file = file;
     this.col = col;
     this.row = row;
+    this.endCol = endCol;
+    this.endRow = endRow;
   }
 
   public Location() {}
@@ -48,12 +59,21 @@ public class Location implements LocationStmt {
     return this;
   }
 
+  public int getEndCol() {
+    return endCol;
+  }
+
+  public int getEndRow() {
+    return endRow;
+  }
+
   @Override
-  public Location setLocation(int file, int row, int col) {
+  public void setLocation(int file, int row, int col, int endRow, int endCol) {
     this.file = file;
     this.col = col;
     this.row = row;
-    return this;
+    this.endCol = endCol;
+    this.endRow = endRow;
   }
 
   @Override
--- a/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/Location.java
+++ b/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/Location.java
@@ -83,22 +83,27 @@ public class Location implements LocationStmt {
 
   @Override
   public boolean equals(Object o) {
-      if (this == o) {
-          return true;
-      }
-      if (o == null || getClass() != o.getClass()) {
-          return false;
-      }
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
 
     Location location = (Location) o;
-
-      if (file != location.file) {
-          return false;
-      }
-      if (col != location.col) {
-          return false;
-      }
-    return row == location.row;
+    if (file != location.file) {
+      return false;
+    }
+    if (col != location.col) {
+      return false;
+    }
+    if (row != location.row) {
+      return false;
+    }
+    if (endCol != location.endCol) {
+      return false;
+    }
+    return endRow == location.endRow;
   }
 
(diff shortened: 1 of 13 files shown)


charlieegan3:
For comparison, this is what we get from the topdown reporting for the same file.

Screenshot

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/stmts/BaseStmt.java
```diff
@@ -8,32 +8,36 @@ public abstract class BaseStmt implements Stmt {
...
+    private int endCol; // column one past the last rune of the source text
+    private int endRow; // row of the last rune of the source text
     private Location cachedLocation;
 
     protected BaseStmt(int file, int col, int row) {
         this.file = file;
         this.col = col;
         this.row = row;
+        this.endCol = col;
+        this.endRow = row;
```
sspaink, line 20: Could you bump the OPA pin in `tools/generate-compliance-tests/go.mod:6` (still `v1.19.0`) to main and re-vendor `opa-proto/src/main/proto/v1/ir/plan.proto`? Upstream now has `end_col = 38`/`end_row = 39` on `message Stmt`, so with that in place `PlanMapper.java:277-279` can map the real end positions and the regenerated fixtures will actually exercise ranges — as it stands the mapper only sets file/col/row, which leaves `endRow`/`endCol` at `0`, inverts the range, and makes `getCoveredLines()` return an empty row set for every proto-format plan.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/Location.java
```diff
@@ -48,12 +59,21 @@ public Location setRow(int row) {
...
+
   @Override
-  public Location setLocation(int file, int row, int col) {
+  public void setLocation(int file, int row, int col, int endRow, int endCol) {
     this.file = file;
     this.col = col;
     this.row = row;
-    return this;
+    this.endCol = endCol;
+    this.endRow = endRow;
```
sspaink, line 76: `endRow`/`endCol` are absent from `equals`/`hashCode`/`toString`, so two Locations describing different ranges compare equal and collide in a set — which also means `PolicyTest`'s round-trip assertions can't catch an end-location wiring error (a swapped `endRow`/`endCol` passes every existing test). `BaseStmt.equals` has the same gap.

```java
if (endCol != location.endCol) {
    return false;
}
return endRow == location.endRow;
```

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/tracing/Range.java
```diff
@@ -0,0 +1,78 @@
...
+    return start;
+  }
+
+  public Position getEnd() {
+    return end;
+  }
+
+  // Mirrors Range.contains in OPA's v1/cover: other is fully contained when this range starts at
+  // or before it and ends at or after it (row compared before column, via Position.compareTo).
+  public boolean contains(Range other) {
```
sspaink: `contains` has no caller outside `RangeTest` — can this drop `public`?

```suggestion
  boolean contains(Range other) {
```

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/stmts/BaseStmt.java
charlieegan3: Thanks, I'd missed that since it's unreleased. https://github.com/open-policy-agent/java-opa-sdk/pull/201/commits/acdd2dabe385c8efda3416fa275045a6ccc81cd4 updates it.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/Location.java
charlieegan3: 4017088 adds this so end values are included.

charlieegan3:
The failing tests are all relating to the OPA update and new examples in the compliance suite. Will do in another PR.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/stmts/BaseStmt.java
charlieegan3: I think we can merge https://github.com/open-policy-agent/java-opa-sdk/pull/203 first since there are some test failures following the tightening up in https://github.com/open-policy-agent/opa/pull/8980

charlieegan3: Then we can keep this PR about the location changes.

charlieegan3:
https://github.com/open-policy-agent/java-opa-sdk/pull/203 is now merged, so this should be more simple!

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/ir/stmts/BaseStmt.java
sspaink: Good idea, thank you!

sspaink (approved):

charlieegan3:
Thanks Sebastian!

PR #202, opened 2026-08-18
time: Match go time format for all jvm locales

charlieegan3:
./gradlew test will fail for non en_US otherwise.


sspaink (approved):
Thanks!

PR #209, opened 2026-08-26
coverage: Add support for unplanned_rules 

charlieegan3:
After https://github.com/open-policy-agent/java-opa-sdk/pull/208

This gives the SDK the capability to load and expose the unplanned_rules
data in the plan if present.


sspaink (approved):
Thank you!

Issue #88, opened 2026-06-23
Relax Checkstyle MethodName for test sources


sspaink:
383 of the 399 `MethodName` warnings are JUnit tests using the widely-adopted `methodName_describesScenario` convention, e.g.

- `findPropertyName_readsSerializedNameOnField`
- `findPropertyName_ignoresAnnotationOnGetter`
- `findPropertyName_returnsNullWhenUnannotated`

The default Checkstyle pattern `^[a-z][a-zA-Z0-9]*$` rejects underscores. Renaming 383 tests would lose readable scenario labels for no benefit.

## Proposed fix

Configure a separate, more permissive `MethodName` pattern for test sources. Simplest path is per-task `configFile` in the root `build.gradle.kts`:

\`\`\`kotlin
tasks.named<Checkstyle>("checkstyleTest") {
    configFile = rootProject.file("config/checkstyle/checkstyle-test.xml")
}
\`\`\`

with `checkstyle-test.xml` overriding only `MethodName`:

\`\`\`xml
<module name="MethodName">
    <property name="format" value="^[a-z][a-zA-Z0-9_]*$"/>
</module>
\`\`\`

This drops the count from 399 to 16 (the production offenders tracked in the rename issue) and means the test-naming convention is allowed by policy, not just by neglect.

polachandu:
Hi @sspaink, Happy to pick if it is free? Please assign it to me.

Issue #56, opened 2026-06-22
Make SimpleMetrics thread-safe


sspaink:
`SimpleMetrics` stores timers in a plain `TreeMap` and the inline Timer has unsynchronized `start`/`stop`/`value`. `Engine.Builder.withMetrics(...)` lets callers pass in a shared instance, so this is a footgun — and the Go OPA `metrics` package locks per-timer plus the outer map (see https://github.com/open-policy-agent/opa/blob/main/v1/metrics/metrics.go), so the precedent is to be safe by default.

* Guard the timer/counter/histogram maps (e.g., `ConcurrentHashMap`).
* Synchronize `Timer.start`/`stop`/`value` (or use `AtomicLong` for the accumulated value).
* Use `AtomicLong` / `LongAdder` for `Counter`.
* Add a concurrency test that exercises a shared `Metrics` from multiple threads.

c879873067877881111:
Measured on macOS 14.8.4, Apple Silicon, 8 cores, OpenJDK 17.0.19.

I have this working locally, but I ran into something the second bullet doesn't cover and I'd rather agree on the semantics before opening a PR.

**Adding `synchronized` is not sufficient — `value()` can still return a negative `Duration`.**

`SimpleTimer` reports `end - start`, and with a shared timer those two fields can come from different threads. Thread A runs `start()` then `stop()`; thread B runs `start()` before A reaches `value()`. Now `end` is A's stop timestamp and `start` is B's later start timestamp, so the subtraction goes negative. Locking makes the reads coherent, it does not make them come from the same interval.

I measured it. I applied exactly what the bullets ask for — timer map to `ConcurrentHashMap` with `computeIfAbsent`, and `start`/`stop`/`value` all `synchronized`, keeping `end - start` — then ran 8 threads × 500 iterations of `start(); stop(); value();` against one shared timer, ten times:

```
RUN  1: FAIL  timer reported a negative duration: PT-0.000000209S
RUN  2: FAIL  timer reported a negative duration: PT-0.000001084S
RUN  3: FAIL  timer reported a negative duration: PT-0.000000208S
RUN  4: FAIL  timer reported a negative duration: PT-0.000000333S
RUN  5: FAIL  timer reported a negative duration: PT-0.000000834S
RUN  6: FAIL  timer reported a negative duration: PT-0.000000750S
RUN  7: FAIL  timer reported a negative duration: PT-0.000000625S
[...]

PR #216, opened 2026-09-03
fix(metrics): implement SimpleMetrics counter and histogram

### LINKED ISSUE #60: SimpleMetrics.histogram() and counter() return null
`SimpleMetrics` only implements `Timer`. `histogram(String)` and `counter(String)` return `null`, so any caller using those metric types NPEs. Implement both with java-idiomatic types, keeping the well-known metric names so decision-log output lines up with OPA's CLI.

* Implement `Counter` (`AtomicLong` / `LongAdder` for `add` and `incr`).
* Implement `Histogram` with min/max/mean/stddev/median + configurable percentiles.
* Backfill unit tests covering all three metric types.

c879873067877881111:
Fixes #60

### Summary

`SimpleMetrics` only implemented `Timer`. `histogram(String)` and `counter(String)`
returned `null`, so any caller of those metric types threw an NPE, and `all()`
returned only the timer map, so neither type ever reached `MetricsPrinter`.

### Changes

- `Counter` backed by an `AtomicInteger`. The issue suggests `AtomicLong`/`LongAdder`,
  but `Metrics.Counter` declares `add(int)` and `int value()`, so a long accumulator
  would only add a lossy narrowing at the boundary.
- `Histogram` over a bounded reservoir of 1028 samples — the size Go-OPA takes from
  `rcrowley/go-metrics` — replacing a random slot once full, so a long-lived `Metrics`
  instance cannot grow without bound. Statistics follow the Go implementation:
  population standard deviation (`SampleVariance` divides by `N`) and the linearly
  interpolated `SamplePercentiles`, at Go's hardcoded set (`median`, `75%`, `90%`,
  `95%`, `99%`, `99.9%`, `99.99%`). `count` is the number of updates, not the reservoir
  occupancy, matching `snap.Count()`.
- `all()` and `clear()` now cover all three metric types.
- `Histogram.Values` keeps its `int` fields, so the rows `MetricsPrinterTest` asserts
  are unchanged.
- The timer map and the inline `Timer` are untouched — that is #56. The new
  counter/histogram code is thread-safe as written.

### Verification

- `./gradlew :opa-evaluator:test :opa-evaluator:checkstyleMain :opa-evaluator:checkstyleTest
  :opa-evaluator:pmdMain :opa-evaluator:pmdTest` — no PMD findings, and no Checkstyle
  findings for `SimpleMetrics.java`. `SimpleMetricsTest` picks up the usual `MethodName`
  warnings for `snake_case` test names, as the existing test sources do (#88).
- Adds `SimpleMetricsTest`, 8 cases. Reverting only `SimpleMetrics.java` fails 7 of the
  8; the 8th covers `clear()`, which passes trivially when nothing can be registered.
- `./gradlew build`: the only remaining failures are `json_match_schema/invalid document`
  and `:opa-proto:vendorProtoSchemas`. The first reproduces unchanged with my change
  reverted; the second is `go mod download` unable to reach `proxy.golang.org` from here.

### Open questions

1. The issue asks for configurable percentiles, but upstream Go is explicitly "a
   histogram with hardcoded percentiles" and CONTRIBUTING asks to keep the public
   surface minimal, so I matched Go. Happy to add a constructor overload if you would
   rather have it configurable.
2. `mean` and `stddev` are `float64` in Go but `int` in `Histogram.Values`, so precision
   is lost. Widening them is a public-API change that also rewrites the
   `MetricsPrinterTest` expectations — worth a separate issue?

### DIFF
--- a/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
+++ b/opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
@@ -1,13 +1,20 @@
 package io.github.open_policy_agent.opa.metrics;
 
 import java.time.Duration;
+import java.util.Arrays;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Timer;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public class SimpleMetrics implements Metrics {
 
   private final Map<String, Timer> timers = new TreeMap<>();
+  private final Map<String, Counter> counters = new ConcurrentHashMap<>();
+  private final Map<String, Histogram> histograms = new ConcurrentHashMap<>();
 
   @Override
   public String name() {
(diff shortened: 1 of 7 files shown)


opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
```diff
@@ -43,21 +50,129 @@ public Duration value() {
...
-    return null;
+    return counters.computeIfAbsent(name, key -> new SimpleCounter());
   }
 
   @Override
   public Map<String, Metric> all() {
-    return new TreeMap<>(timers);
+    Map<String, Metric> all = new TreeMap<>(timers);
+    all.putAll(counters);
+    all.putAll(histograms);
```
sspaink, line 65: Now that counters and histograms are in `all()`, `DecisionLogPlugin.buildDecisionEvent` falls into its `else` branch and calls `MAPPER.valueToTree(...)` on `SimpleCounter`/`SimpleHistogram` — neither has a Jackson-visible property, so it throws `No serializer found ... and no properties discovered to create BeanSerializer`. `DecisionLogs.logDecision` swallows that as `"Failed to log decision"`, so the entire decision event is dropped, which is the opposite of the CHANGELOG's "so they reach ... decision logs".

`MetricsModule` already solves this for `Timer`; the same mixin treatment covers the other two:

```java
setMixInAnnotation(Metrics.Counter.class, CounterMixin.class);
setMixInAnnotation(Metrics.Histogram.class, HistogramMixin.class);

abstract static class CounterMixin {
  @JsonValue
  abstract int value();
}

abstract static class HistogramMixin {
  @JsonValue
  abstract Metrics.Histogram.Values value();
}
```

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
```diff
@@ -43,21 +50,129 @@ public Duration value() {
...
+      if (pos >= sorted.length) {
+        return sorted[sorted.length - 1];
+      }
+      double lower = sorted[(int) pos - 1];
+      double upper = sorted[(int) pos];
+      return lower + (pos - Math.floor(pos)) * (upper - lower);
+    }
+
+    private static int round(double value) {
+      return (int) Math.round(value);
```
sspaink: This can overflow. The largest value an `int` holds is about 2.1 billion, which is only 2.15 seconds in nanoseconds — and nanosecond timings are what Go feeds histograms in `topdown/instrumentation.go`. Anything bigger turns negative: `update(3e9)` reports `-1294967296` for every stat. The line above does `Math.min(count, Integer.MAX_VALUE)` to avoid this; the same idea works here:

```suggestion
      return (int) Math.max(Integer.MIN_VALUE, Math.min(Integer.MAX_VALUE, Math.round(value)));
```

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
```diff
@@ -43,21 +50,129 @@ public Duration value() {
...
   public Counter counter(String name) {
-    return null;
+    return counters.computeIfAbsent(name, key -> new SimpleCounter());
   }
 
   @Override
   public Map<String, Metric> all() {
-    return new TreeMap<>(timers);
+    Map<String, Metric> all = new TreeMap<>(timers);
+    all.putAll(counters);
```
sspaink, line 64: Since all three maps share one key space, reusing a name across types silently drops one. After `metrics.timer("x")` and `metrics.counter("x")`, `all()` returns just the counter — the timer is gone from the printer, the CLI table and decision logs. Go's `All()` prefixes each key by type (`timer_x`, `counter_x`), so both survive there.

Matching that means moving the prefixing out of `MetricsPrinter` and into `all()`, which is bigger than this PR — fine as a follow-up, but worth noting the gap here.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
c879873067877881111: Mixins added. Testing them turned up a second path: `DecisionLogs.MAPPER` is a bare `ObjectMapper + JavaTimeModule,` and `opa-jackson` is only `runtimeOnly` for `opa-services`, so the mixins don't reach it — with them in place the repro still failed identically. Timers avoid it only because of the explicit `timer_<name>_ns branch`.
So `buildDecisionEvent` now writes counters and histograms itself as well, under `formatKey's` names, `counter_<name>` and `histogram_<name>`; unknown Metrics keep the Jackson fallback. Covered by `decisionLogs_includesEveryMetricType`.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
c879873067877881111: Agreed with all three types sharing a name, only one survives `all()`. 
Confirmed `formatKey` behaves as you describe. Leaving it out of this PR
Do you want to open an issue to track it?

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
c879873067877881111: The suggestion was adopted without modification.
Overflow occurs in the negative direction as well—`update(-3e9)` results in `+1294967296`; his original comment only mentioned the positive-to-negative case.
Both directions were tested; the test names are `histogram_saturatesStatsAboveIntRange` and `histogram_saturatesStatsBelowIntRange`.
I left Values as int rather than widening to Go's int64/float64.

sspaink (approved):
thanks!

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/metrics/SimpleMetrics.java
sspaink: issue opened: https://github.com/open-policy-agent/java-opa-sdk/issues/229)

PR #220, opened 2026-09-08
coverage: Report not_covered lines

charlieegan3:
Walk every statement in the IR plan to find ranges that were compiled but never executed in a given run.

This brings the Java implementation here in line with OPA's v1/cover, which reports such gaps in this way. 'Unplanned Rules' are not really a thing in the go path, so this library brings these together to match the OPA output, if not the method for unplanned and thus discarded rules.


sspaink (approved):
Thank you!

charlieegan3:
Thanks for the review!

Issue #64, opened 2026-06-22
Support OPA delta bundles


sspaink:
OPA's bundle protocol supports delta bundles for incremental data updates: a `patch.json` describing add/replace/remove operations against the existing data tree, instead of a full replacement. The Java SDK currently treats every bundle as a full replacement, which scales poorly for large data sets that change frequently.

* Detect delta bundles via the manifest (`roots` + delta indicators) and parse `patch.json`.
* Apply patches to the in-store data tree without re-loading policies.
* Reject delta bundles that violate the active manifest's `roots` (depends on #63).
* Tests for add / replace / remove ops and for invalid patches.

Issue #39, opened 2026-05-27
Support bundle signature verification


johanfylling:
Add support for OPA's signed bundle format (see [docs](https://www.openpolicyagent.org/docs/management-bundles#signing)).

* Config: add top-level keys block and signing section under `BundleConfig` (keyid, scope, exclude_files).
* `TarballBundleLoader`: extract `.signatures.json` and compute per-file digests. 
* New verifier: validate JWS over file digests using configured public key/secret.
* `BundlePlugin`: fail activation when verification fails.

PR #226, opened 2026-09-20
feat(bundle): replace manifest map with typed Manifest

### LINKED ISSUE #69: Replace Map<String,Object> manifest with typed model
`Bundle.manifest` is `Map<String,Object>`, parsed from `.manifest` by the `BundleParser` SPI. Every consumer (manifest enforcement, status events, decision logs) has to navigate the untyped tree and re-validate field types. A typed model would centralize that and let the compiler catch field-name typos.

* Define a `Manifest` value type with `revision`, `roots`, `metadata`, and a place for delta indicators (#64) and signing fields (#39).
* Have `BundleParser` produce `Manifest` directly; keep the raw map only if needed for round-tripping.
* Update `Bundle`, `BundlePlugin`, and downstream consumers.
* Should land before — or alongside — #63, #64, #68 to avoid each one inventing its own ad-hoc accessors.

CD57:
Fixes #69

Bundle manifests were exposed as Map<String, Object>. Every place that read it (ie  store, provenance, status, decision-log plugins) had to know the field names, check the types, and convert the values itself.

PR adds a Manifest value type to opa-evaluator and has BundleParser and ProtoBundleDecoder return it, so the manifest is validated once at load and then used via accessors. The fields mirror bundle.Manifest in OPA, defaults coming from Manifest.Init and unknown fields kept and accessed via asMap(). Breaking change for anyone with a custom parser, which now has to return a Manifest, and for callers doing map lookups. Manifests with the wrong field types (ie a numeric revision) are now rejected at load, which is what OPA does.

Status reports and decision logs now leave out an empty revision, since Status.ActiveRevision and BundleInfoV1.Revision are both omitempty, and provenance lists every bundle with "" when no revision is recorded, as OPA's getRevisions does.

The issue also asks for a place to put delta indicators (#64) and signing fields (#39). I looked into it and in OPA those aren't part of the manifest as Patch and Signatures live on Bundle, populated from patch.json and .signatures.json, and a bundle is treated as a delta simply by having a patch.json. So I've left them out of Manifest for now rather than guess at a shape. Happy to add them here if you'd prefer, or leave that for #64 and #39.

Please review and let me know if any changes are required, thank you.

OPA references (v1.20.1):
[bundle.Manifest](https://github.com/open-policy-agent/opa/blob/v1.20.1/v1/bundle/bundle.go#L147) for the field set
[Manifest.Init](https://github.com/open-policy-agent/opa/blob/v1.20.1/v1/bundle/bundle.go#L181) for defaulting missing roots to [""]
[Status.ActiveRevision](https://github.com/open-policy-agent/opa/blob/v1.20.1/v1/plugins/bundle/status.go#L27) for omitting an empty revision from status reports
[BundleInfoV1.Revision](https://github.com/open-policy-agent/opa/blob/v1.20.1/v1/plugins/logs/plugin.go#L80) for omitting an empty revision from decision logs
[getRevisions](https://github.com/open-policy-agent/opa/blob/v1.20.1/v1/server/server.go#L1058) for listing every bundle in provenance with "" when no revision is recorded
[Patch](https://github.com/open-policy-agent/opa/blob/v1.20.1/v1/bundle/bundle.go#L71) and [Signatures](https://github.com/open-policy-agent/opa/blob/v1.20.1/v1/bundle/bundle.go#L64) on Bundle, not Manifest
[Bundle.Type()](https://github.com/open-policy-agent/opa/blob/v1.20.1/v1/bundle/bundle.go#L1494) for classifying a bundle as delta by its patch.json


sspaink (commented):
Nice cleanup — the typed model reads much better than the raw map, and the revision semantics in status/decision logs line up with OPA. Three notes, the first being the one I'd want resolved before merge.

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/bundle/Manifest.java
```diff
@@ -0,0 +1,336 @@
...
+      throw wrongType(field, "integer", value);
+    }
+    try {
+      return value instanceof BigInteger integer
+          ? integer.intValueExact()
+          : Math.toIntExact(((Number) value).longValue());
+    } catch (ArithmeticException e) {
+      throw wrongType(field, "32-bit integer", value);
+    }
+  }
```
sspaink: Parsers other than Jackson decode `"rego_version": 1` as `1.0`, so a Gson/org.json-backed `BundleParser` now fails on every `opa build` bundle with `must be integer, got Double`.

```suggestion
  private static Integer requireInteger(String field, Object value) {
    if (!(value instanceof Number number)) {
      throw wrongType(field, "integer", value);
    }
    try {
      // Parsers other than Jackson decode JSON `1` as `1.0`, so accept integral floating-point too.
      return new BigDecimal(number.toString()).intValueExact();
    } catch (ArithmeticException | NumberFormatException e) {
      throw wrongType(field, "32-bit integer", value);
    }
  }
```

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/bundle/Manifest.java
```diff
@@ -0,0 +1,336 @@
...
+    if (!(other instanceof Manifest that)) {
+      return false;
+    }
+    return properties.equals(that.properties);
+  }
+
+  @Override
+  public int hashCode() {
+    return properties.hashCode();
+  }
```
sspaink: Raw-map equality makes the same bundle unequal across formats — the proto path always emits `revision` (faithfully; Go has no `omitempty` there), a hand-written `{"roots":["x"]}` manifest doesn't. `ProtoBundleParityTest.decodedManifestMatchesJsonManifest` passes only because the fixture spells out `"revision":""`.

```suggestion
  @Override
  public boolean equals(Object other) {
    if (this == other) {
      return true;
    }
    if (!(other instanceof Manifest that)) {
      return false;
    }
    return revision.equals(that.revision)
        && roots.equals(that.roots)
        && wasm.equals(that.wasm)
        && Objects.equals(regoVersion, that.regoVersion)
        && fileRegoVersions.equals(that.fileRegoVersions)
        && Objects.equals(metadata, that.metadata)
        && Objects.equals(defaultDecision, that.defaultDecision)
        && additionalProperties.equals(that.additionalProperties);
  }

  @Override
  public int hashCode() {
    return Objects.hash(revision, roots, wasm, regoVersion, fileRegoVersions, metadata, defaultDecision, additionalProperties);
  }
```

opa-evaluator/src/main/java/io/github/open_policy_agent/opa/bundle/Manifest.java
```diff
@@ -0,0 +1,336 @@
...
+        || value instanceof Long
+        || value instanceof BigInteger
+        || value instanceof BigDecimal
+        || (value instanceof Float number && Float.isFinite(number))
+        || (value instanceof Double number && Double.isFinite(number))
+        || value instanceof Boolean) {
+      return value;
+    }
+    throw new IllegalArgumentException(
+        "Manifest values must use JSON-compatible types, got " + value.getClass().getSimpleName());
```
sspaink: A `NaN` in `metadata` — which Go OPA loads fine, since `Metadata` is `map[string]any` and never read — fails the whole bundle with "must use JSON-compatible types, got Double", pointing at the type rather than the value.

```suggestion
    if (value instanceof Float || value instanceof Double) {
      throw new IllegalArgumentException("Manifest values must be finite numbers, got " + value);
    }
    throw new IllegalArgumentException(
        "Manifest values must use JSON-compatible types, got " + value.getClass().getSimpleName());
```

